PR #381, opened 2020-04-09
Submitted models/robotika kloubak sensor config 1

nkoenig:
* Model submission for Kloubak sensor configuration 1 from Robotika
* Kloubak URDF ver0


acschang:
The initial assessment of the Robotika Kloubak model is complete, however there are some issues to resolve before this model can be merged and utilized in a competitive setting:

- The returns from the LIDAR do not appear to be accurate; please correct the orientation of the mounted LIDAR.
- The odometry for both the front and rear wheels is published on the same topic and conflict with one another. The odometry will need to be split to two separate topics as to prevent this conflict.
- The battery life of the vehicle is 3 minutes instead of the specified time of 5400 seconds. Please resolve.
- This model has a non-standard setup with two velocity controllers. In the "Control" section of `specifications.md`, please describe how the two velocity controllers are intended to be used. For example, should both topics `cmd_vel/front` and `cmd_vel/rear` be commanded with positive velocities to move the vehicle forward?
- The validation data is necessary to complete the evaluation. If not provided, the model's maximum velocity will be limited to 1m/s and battery life set to 60min. When the model validation data can be provided later, please submit a PR updating the model.

Please address the issues noted above and then submit a pull request to update your submission by **5 JUNE**. The submissions are expected to be incorporated into the SubT Virtual Testbed pending a successful review.

m3d:
The first batch of fixes is available here https://github.com/osrf/subt/pull/445

acschang:
@m3d Given the unique characteristics of the Kloubak model with a joint between the front and rear differential drives, we believe outlining your standard operating procedures for commanding the vehicle would be helpful for competitors interested in using your model. For example, are you supposed to command both halves to move the robot forward, or only one at a time? What about turning and accounting for the possibility of [jackknifing](https://en.wikipedia.org/wiki/Jackknifing)?

acschang:
`DiffDrive` odometry topic specification is dependent upon [this PR](https://github.com/ignitionrobotics/ign-gazebo/pull/179).

m3d:
> @m3d Given the unique characteristics of the Kloubak model with a joint between the front and rear differential drives, we believe outlining your standard operating procedures for commanding the vehicle would be helpful for competitors interested in using your model. For example, are you supposed to command both halves to move the robot forward, or only one at a time? What about turning and accounting for the possibility of [jackknifing](https://en.wikipedia.org/wiki/Jackknifing)?

@acschang - thanks for corrections. I can only describe how is controlled the real robot (see
https://robotika.cz/robots/kloubak/cs (unfortunately only in Czech)). Each wheel has BLDC motor with Hall sensors and dedicated motor controller (i.e. 4 motor controllers in total). They can keep speed, brake or leave the "free wheel". When we used Kloubak K2 in Pittsburgh for the first time the full control was not finished yet, but it turned out to be advantage. Only two the front wheels were driven and rear wheels were passive and provide very precise odometry (we are not able to model this well, what reported @jisa). The robot was symmetric so instead of turning you could virtually flip front & rear and drive original rear wheels to return back.

Jakub (@tajgr) implemented 4WD for Urban Circuit in Olympia and also the tires were upgraded. You can switch on demand to control all four wheels but you pay by loosing precise odometry from passive wheels.

Kloubak K2 has only passive joint with sensor to measure the angle. There are HW limits so that front and rear wheels do not touch (internally there are also SW limits to avoid reaching HW limits). The 4WD controller SW takes the joint angle as one of the inputs and tries to maintain desired joint angle and vehicle velocity.

m3d:
p.s. there is small update with specification https://github.com/osrf/subt/pull/449

nkoenig:
I've updated this model to use the meshes from https://sketchfab.com/3d-models/robotika-kloubak-k2-f570c19557924e7c91ee374ce41b7c16. Here is the model https://app.ignitionrobotics.org/OpenRobotics/fuel/models/Robotika_Kloubak_Sensor_Config_1

m3d:
@nkoenig thanks - the new models/meshes look cool! :)

nkoenig:
I'm merging this PR into the `cave_feature_release3` branch. This robot will be available when the feature release branch is merged into master.

PR #392, opened 2020-04-14
Cave prep

nkoenig:
This is mostly code cleanup after moving from bitbucket.

### DIFF
--- a/docker/build.bash
+++ b/docker/build.bash
@@ -37,14 +37,10 @@ fi
 user_id=$(id -u)
 image_name=$(basename $1)
 image_plus_tag=$image_name:$(date +%Y_%b_%d_%H%M)
-# mercurial adds a + symbol if there are uncomitted changes in the repo
-# that will break docker tag syntax
-hg_id=$(hg id -i | tr -d '+')
 
 shift
 
 docker build --rm -t $image_plus_tag --build-arg user_id=$user_id "$@" -f $DIR/$image_name/Dockerfile .
 docker tag $image_plus_tag $image_name:latest
-docker tag $image_plus_tag $image_name:$hg_id
 
-echo "Built $image_plus_tag and tagged as $image_name:latest and $image_name:$hg_id"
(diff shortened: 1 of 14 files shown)


docker/cloudsim_sim/Dockerfile
```diff
@@ -108,7 +107,7 @@ RUN rosdep update
 # docker is run
 RUN mkdir -p subt_ws/src \
  && cd subt_ws/src \
- && hg clone https://bitbucket.org/osrf/subt
+ && git clone git@github.com:osrf/subt
```
iche033, line 110: use https? this requires ssh keys to be set up on github

iche033: same goes for `subt_shell` and `subt_sim_entry`

docker/build.bash
```diff
@@ -37,14 +37,10 @@ fi
 user_id=$(id -u)
 image_name=$(basename $1)
 image_plus_tag=$image_name:$(date +%Y_%b_%d_%H%M)
-# mercurial adds a + symbol if there are uncomitted changes in the repo
-# that will break docker tag syntax
-hg_id=$(hg id -i | tr -d '+')
```
mjcarroll, line 42: When I ported this, I ended up using:

`git_id=$(git describe)` 

It requires you to create annotated tags with `git tag -a` and make sure they are pushed with `git push --tags`

mjcarroll: And it generates version numbers like `0.0.1-14-g136b05b` which is `$(last tag)-$(commits since tag)-g$(commit hash)`

nkoenig: Thanks. I removed it because we don't use these tagged images, so it felt like a useless chunk of code.

docker/cloudsim_sim/Dockerfile
nkoenig: Changed in  0a4d7d0be79a550403bd6718a14889773090455e

PR #398, opened 2020-04-16
Adding breadcrumbs to models

nkoenig:
## Models Added

1. X1_SENSOR_CONFIG_7 (420 credits) - X1C4 with 12 breadcrumbs
1. X2_SENSOR_CONFIG_8 (275 credits) - X2C6 with 6 breadcrumbs
1. COSTAR_HUSKY_SENSOR_CONFIG_2 (385 credits) - COSTAR Husky Config 1 with 12 breadcrumbs 
1. EXPLORER_X1_SENSOR_CONFIG_2 (440 credits) - EXPLORER X1 with 12 breadcrumbs
1. X1_SENSOR_CONFIG_8 (370 credits) - X1C3 with 12 breadcrumbs
1. X2_SENSOR_CONFIG_9 (205 credits) - X2C4 with 6 breadcrumbs

To test:
```
ign launch -v 4 cave_circuit.ign worldName:=simple_cave_01 robotName1:=X1 robotConfig1:=X1_SENSOR_CONFIG_7
```

To drop a breadcrumb
```
ign topic -t "/model/X1/breadcrumb/deploy" -m ignition.msgs.Empty -p "unused: true"
```


submitted_models/costar_husky_sensor_config_2/launch/spawner.rb
```diff
@@ -0,0 +1,85 @@
...
+        <type>gas</type>
+      </plugin>
+      <plugin filename="libignition-gazebo-breadcrumbs-system.so"
+        name="ignition::gazebo::systems::Breadcrumbs">
+        <topic>/model/#{_name}/breadcrumb/deploy</topic>
+        <max_deployments>12</max_deployments>"
+        <breadcrumb>"
+          <sdf version="1.6">
+            <model name="#{_name}__breadcrumb__">
+              <pose>-1.2 0 0 0 0 0</pose>
```
azeey, line 62: After testing these out for a little bit, a 1.2 m offset seems a little too far from the robot. If the robot's rear is right up against a wall, deploying the breadcrumb could put the breadcrumb inside the wall, or worse, outside the cave. It could also be a problem if there is another robot right behind a robot deploying a breadcrumb.

nkoenig: Would it be possible to make the breadcrumb model not collide with the robot model, and spawn the breadcrumb inside the robot?

azeey: The Collision bitmasks feature was added in ign-physics2, so we can try to backport it to ign-physics1. With that, we can make breadcrumbs not collide with any robot. I don't think we can easily make it so that it wouldn't collide only with the robot that spawned it. 

Would decreasing the offset not work?

nkoenig:
Version 2 of the breadcrumb model uses a box collision shape: https://app.ignitionrobotics.org/openrobotics/fuel/models/Breadcrumb%20Node

submitted_models/costar_husky_sensor_config_2/launch/spawner.rb
nkoenig: I was just thinking creatively. I'll decrease the offset.

nkoenig:
This PR prevents the deployment topic from being created: https://github.com/ignitionrobotics/ign-gazebo/pull/88

azeey:
I guess my concern was more on the ROS side because that's the API for competitors. IIUC, the ROS topic will still exist even if there is no corresponding topic in ign-transport.

nkoenig:
I blocked the ROS topics here: 5f7697a091b9208ff0932a88b6343b61f1984762

azeey:
Looks great! I think we also need to to add the same logic in `cloudsim_bridge.ign`

nkoenig:
Thanks, added: fe173a8139a8c5d616da275b7cbe61c31e11fd16

submitted_models/costar_husky_sensor_config_2/launch/spawner.rb
nkoenig: Adjusted spawn location: a1a886b42f913692dfb8fe9d099ef21ef4f22586

nkoenig:
Thanks, I've applied those changes.

PR #404, opened 2020-04-20
Comms visualization citadel

nkoenig:
This will mostly work with ign-citadel. A new release of ign-gazebo3 is needed.


subt_ign/src/LogChecker.cc
```diff
@@ -193,7 +193,7 @@ class LogChecker
       return 1;
 
     // Begin playback as fast as possible.
-    const auto handle = this->player->Start(std::chrono::seconds(1), false);
+    const auto handle = this->player->Start(std::chrono::seconds(1));
```
iche033, line 196: just checking if the removal of second arg is intended? Looking at [ign-transport doc](https://github.com/ignitionrobotics/ign-transport/blob/ign-transport8/log/include/ignition/transport/log/Playback.hh#L96) the default is `true` which means it will no longer playback as fast as possible. If that's intended, we can just update the comment above this line.

caguero:
I can't compile this branch using Citadel. It seems that `subt_ign/src/VisibilityPlugin.cc` is using the `AxisAlignedBox` component. However [this component doesn't exist anymore in ign-gazebo 3](https://github.com/ignitionrobotics/ign-gazebo/tree/ign-gazebo3/include/ignition/gazebo/components).

iche033:
looks like ign-gazebo2 needs to be merged forward to ign-gazebo3 to get the AxisAlignedBox changes

iche033:
ported axis aligned box changes forward to ign-gazebo3 in ignitionrobotics/ign-gazebo#127

iche033:
to test comms visualization:

```
# launch cave world with X1:
ign launch -v 4 cave_circuit.ign worldName:=simple_cave_01 robotName1:=X1 robotConfig1:=X1_SENSOR_CONFIG_1

# make service call
ign service -s /subt/comms_model/visualize --reqtype ignition.msgs.StringMsg --reptype ignition.msgs.Boolean --timeout 2000 --req 'data: "X1"'


```

nkoenig:
works for me as well.

Issue #90, opened 2019-07-01
Physics issues with Ignition simulation


osrf-migration:
**Original report ([archived issue](https://osrf-migration.github.io/subt-gh-pages/#!/osrf/subt/issues/90)) by Jon Fink (Bitbucket: [jonfink-arl](https://bitbucket.org/%7B89cf7350-7d3a-4a2e-a6ce-43d49bfe9001%7D/)).**

The original report had attachments: [X1behavior.mp4](https://osrf-migration.github.io/subt-gh-pages/data/repositories/osrf/subt/issues/90/attachments/X1behavior.mp4)

----------------------------------------
Driving the X1 robot through the practice tunnel environment \(launched via `ign launch -v 4 competition.ign robotName1:=X1 robotConfig1:=X1_SENSOR_CONFIG_1`, it seems that some of the friction properties for the tunnel may be incorrect after migrating to Ignition with DART \(I believe\) as the default physics engine. 

I can drive around the staging area just fine. Proceeding into the tunnel things seem ok until I start going down the first decline. At this point the robot goes _very_ fast \(no matter what velocity input\). After eventually coming to a stop, I notice that I cannot turn the robot. 

**It seems like the friction settings for the tunnel surface are much to low**

PR #410, opened 2020-04-30
Make breadcrumb model static after specified time

iche033:
depends on [ignition-gazebo pull request #90](https://github.com/ignitionrobotics/ign-gazebo/pull/90)

Add `disable_physics_time` param to breadcrumb plugin that turns the breadcrumb model static after ~3 sim time seconds. A static entity is created when the model becomes static, and that indicates to the comms broker that it should update its visibility graph.

You should see output similar to this:

```
[Msg] Deploying X1__breadcrumb___0 at 3.55 5 0.13228 0 0 0
[Dbg] [EntityComponentManager.cc:627] Using components of type [285520019272511910] / [ign_gazebo_components.DetachableJoint].
[Msg] Breadcrumb 'X1__breadcrumb___0' is now static.
[Dbg] [Physics.cc:864] Creating detachable joint [729]
[GUI] [Dbg] [EntityComponentManager.cc:627] Using components of type [285520019272511910] / [ign_gazebo_components.DetachableJoint].
[Msg] New breadcrumb detected, visibility graph updated
```

### DIFF
--- a/subt_ign/src/CommsBrokerPlugin.cc
+++ b/subt_ign/src/CommsBrokerPlugin.cc
@@ -19,6 +19,7 @@
 
 #include <functional>
 #include <mutex>
+#include <regex>
 
 #include <ignition/common/Console.hh>
 #include <ignition/common/Util.hh>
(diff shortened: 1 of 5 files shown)


subt_ign/src/CommsBrokerPlugin.cc
```diff
@@ -310,7 +311,9 @@ void CommsBrokerPlugin::UpdateIfNewBreadcrumbs()
   for (const auto& [name, pose] : this->poses)
   {
     // New breadcrumb found.
-    if (name.find("__breadcrumb__") != std::string::npos &&
+    // A static model is spawned when the breadcrumb is made static
+    if (std::regex_match(name,
```
nkoenig, line 315: Confirming for documentation purposes that the breadcrumb will only affect communication after the `<disable_physics_time>`, which is currently 3 seconds, correct?

iche033: yes that's correct

PR #411, opened 2020-05-01
Adding team base

nkoenig:
This adds a `teambase`. The `teambase` cannot move, an is permanently in the staging area. It's specified in the same manner as other robots, but the name and config must be "teambase" or "team_base". Capitalization doesn't matter.

**Example**

```
ign launch cave_circuit.ign worldName:=simple_cave_01 robotName1:=teambase robotConfig1:=teambase
```

You should see a small black box in the staging area. It has no collision object, just a visual.

Only 1 team base is allowed.

### DIFF
--- a/subt_ign/launch/cave_circuit.ign
+++ b/subt_ign/launch/cave_circuit.ign
@@ -1038,6 +1038,47 @@
     "  <command>roslaunch --wait subt_ros vehicle_topics.launch world_name:=#{$worldName} name:=#{_name} uav:=#{uav} laser_scan:=#{laserScan} stereo_cam:=#{stereoCam} rgbd_cam:=#{rgbdCam}</command>\n"\
     "</executable>\n"
   end
+
+  def spawnTeamBase(_x, _y, _z, _yaw)
+    "<plugin name=\"ignition::launch::GazeboFactory\"\n"\
+    "        filename=\"libignition-launch-gazebo-factory.so\">\n"\
+    "  <name>TeamBase</name>\n"\
+    "  <allow_renaming>false</allow_renaming>\n"\
+    "  <world>#{$worldName}</world>\n"\
+    "  <is_performer>false</is_performer>\n"\
+    "  <sdf version='1.6'>\n"\
+    "  <model name='TeamBase'>\n"\
+    "    <pose>#{_x} #{_y} #{_z} 0 0 #{_yaw}</pose>\n"\
+    "    <static>true</static>\n"\
+    "    <link name='link'>\n"\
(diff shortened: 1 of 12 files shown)


pauljurczak:
What is its purpose? Is it documented somewhere?

nkoenig:
A wiki page will come shortly. This gives a team the ability to have a zero-cost computer, the Team Base, in the staging area. You'll be able to assign a docker image to the Team Base, just as you would a regular robot. 

The Team Base cannot move, and you can only have one. This is also optional.

subt_ign/launch/cave_circuit.ign
```diff
@@ -1038,6 +1040,47 @@
...
+    "      <visual name='visual'>\n"\
+    "        <geometry>\n"\
+    "          <box>\n"\
+    "            <size>.1 .1 .1</size>\n"\
+    "          </box>\n"\
+    "        </geometry>\n"\
+    "      </visual>\n"\
+    "    </link>\n"\
+    "    <!-- Publish robot state information -->\n"\
+    "    <plugin filename=\"libignition-gazebo-pose-publisher-system.so\"\n"\
```
iche033, line 1066: I think we don't need to use pose publisher system for this simple model and  instead we can just use a `static_transform_publisher` in `teambase_topics.launch`, which helps to reduce data published to the `/tf` topic

e.g. the `teambase_topics.launch` would then be:

```
<?xml version="1.0"?>
<launch>
  <node pkg="tf2_ros" type="static_transform_publisher" name="tf_teambase_static" args="0 0 0.05 0 0 0 TeamBase TeamBase/link"/>
</launch>
```

teambase_description/urdf/teambase.xacro
```diff
@@ -0,0 +1,13 @@
+<?xml version='1.0' encoding='utf-8'?>
+<robot name="TeamBase" xmlns:xacro="http://ros.org/wiki/xacro">
+  <xacro:property name="robot_namespace" value="TeamBase"/>
+
+  <link name='${robot_namespace}/base_link'>
```
iche033, line 5: change `base_link` to `link` to match the actual link name of the TeamBase model?

subt_ign/launch/cave_circuit.ign
nkoenig: Thanks, this should be fixed in 32252498f9252f310d06120ffc17673e307b2e24

teambase_description/urdf/teambase.xacro
nkoenig: Fixed in 32252498f9252f310d06120ffc17673e307b2e24

PR #416, opened 2020-05-13
Add level generator

iche033:
There are two methods for level generation in the origin tile_tsv.py script: 1) row and col method, and 2) single tile (distance based) method. This PR ports the row and col level generation method to a program called `level_generator`  

The code for parsing sdf file from the `dot_generator` was refactored and reused in the `level_generator`

usage:

```
./level_generator <path_to_sdf> <size_x> <size_y> <size_z> <buffer> <output_file>
```

example:

```
./level_generator simple_cave_01.sdf 25 25 25 5 simple_cave_01.sdf
```

**Note** remove the existing levels before running the `level_generator` to avoid adding duplicate levels to the sdf

I checked the results against the levels in urban practice worlds. The order of `<level>` SDF params and names do not match due to differences in how tiles and artifacts are read and stored. So manually compared the level positions and geometry size and they seem to be correct. One key difference is that the z pos and size of level in the original script are hardcoded but the new `level_generator` computes these z values. This is because the vertical depth of the cave worlds can be much larger than previous circuits..

For testing, you can set [debugLevels](https://github.com/osrf/subt/compare/cave_feature_release1...level_generator?expand=1#diff-0630163bb4dab92293c50c7c7b2397fbR60) to `true` to also print out a list of visuals representing the levels. There will be quite a bit of overlap (and z fighting) between visuals so it may be easier to select and highlight these debug levels by using the `EntityTree` gui plugin.

### DIFF
--- a/subt_ign/src/level_generator.cc
+++ b/subt_ign/src/level_generator.cc
@@ -119,56 +119,32 @@ std::string levelsStr(const std::map<double, Level> &_levels)
 
 /// \brief Print the level sdf elements
 /// \param[in] _vertexData vector of model info
-///\ param[in[ _size Tile size
-///\ param[in[ _buffer Level bfufer size
+/// \param[in] _size Tile size, which is used to compute the size of levels
+/// At minimum this should be the size of largest tile in the world but larger
+/// values can be specified to increase the size of each level.
+/// tile in the world.
+/// \param[in] _buffer Level buffer size
 void printLevels(std::vector<VertexData> &_vertexData,
     const math::Vector3d &_size, double _buffer)
 {
   std::map<double, Level> levelX;
   std::map<double, Level> levelY;
 
-  double minX = std::numeric_limits<double>::max();
-  double minY = minX;
-  double minZ = minX;
-  double maxX = -std::numeric_limits<double>::max();
-  double maxY = maxX;
-  double maxZ = maxX;
+  math::Vector3d min(math::MAX_D, math::MAX_D, math::MAX_D);
+  math::Vector3d max(-math::MAX_D, -math::MAX_D, -math::MAX_D);
 
-  // get min max of xyz of models
+  // get min & max model position values
   for (const auto &vd : _vertexData)
   {
-    double x = vd.model.Pose().Pos().X();
-    double y = vd.model.Pose().Pos().Y();
-    double z = vd.model.Pose().Pos().Z();
-
-    if (x < minX)
-      minX = x;
-    if (y < minY)
-      minY = y;
-    if (z < minZ)
-      minZ = z;
-
-    if (x > maxX)
-      maxX = x;
-    if (y > maxY)
-      maxY = y;
-    if (z > maxZ)
-      maxZ = z;
+    min.Min(vd.model.Pose().Pos());
+    max.Max(vd.model.Pose().Pos());
   }
 
   // compute level size and position
-  double lsx = maxX - minX;
-  double lx = minX + lsx * 0.5;
+  math::Vector3d ls = max - min;
+  math::Vector3d l = min + ls * 0.5;
 
-  double lsy = maxY - minY;
-  double ly = minY + lsy * 0.5;
-
-  double lsz = maxZ - minZ;
-  double lz = minZ + lsz * 0.5;
-
-  lsx += _size.X() * 2.0;
-  lsy += _size.Y() * 2.0;
-  lsz += _size.Z() * 2.0;
+  ls += _size * 2.0;
 
   double levelSizeFactor = 1.875;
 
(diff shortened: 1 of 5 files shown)


subt_ign/src/level_generator.cc
```diff
@@ -0,0 +1,313 @@
...
+
+    out << "      </level>\n";
+  }
+  return out.str();
+}
+
+
+/// \brief Print the level sdf elements
+/// \param[in] _vertexData vector of model info
+///\ param[in[ _size Tile size
```
nkoenig, line 122:
```suggestion
/// \param[in[ _size Tile size
```

subt_ign/src/level_generator.cc
```diff
@@ -0,0 +1,313 @@
...
+///\ param[in[ _buffer Level bfufer size
```
nkoenig, line 123:
```suggestion
/// \param[in[ _buffer Level bfufer size
```

subt_ign/src/level_generator.cc
nkoenig, line 122: Is `_size` the size of each level?

subt_ign/src/level_generator.cc
```diff
@@ -0,0 +1,313 @@
...
+void printLevels(std::vector<VertexData> &_vertexData,
```
nkoenig, line 124: I think you can simplify the computation of `min`, `max`, `ls` and `l`:

```
void printLevels(std::vector<VertexData> &_vertexData,
    const math::Vector3d &_size, double _buffer)
{
  std::map<double, Level> levelX;
  std::map<double, Level> levelY;

  math::Vector3d min(math::MAX_D, math::MAX_D, math::MAX_D);
  math::Vector3d max(-math::MAX_D, -math::MAX_D, -math::MAX_D);

  // get min & max model position values
  for (const auto &vd : _vertexData)
  {
    min.Min(vd.model.Pose().Pos());
    max.Max(vd.model.Pose().Pos());
  }

  // compute level size and position
  math::Vector3d ls = max - min;
  math::Vector3d l = min + ls * 0.5;

  ls += _size * 2.0;

  double levelSizeFactor = 1.875;

  for (const auto &vd : _vertexData)
  {
    std::string name = vd.tileName;

    double x = vd.model.Pose().Pos().X();
    double y = vd.model.Pose().Pos().Y();

    // Create a new level for each unique x
    if (levelX.find(x) == levelX.end())
    {
      double rx = x;
      double ry = l.Y();
      double rz = l.Z();

      double rsx = _size.X() * levelSizeFactor;
      double rsy = ls.Y();
      double rsz = ls.Z();

      Level level;
      level.x = rx;
      level.y = ry;
      level.z = rz;
      level.sx = rsx;
      level.sy = rsy;
      level.sz = rsz;
      level.buffer = _buffer;
      levelX[x] = level;
    }
    levelX[x].refs.push_back(name);

    // Create a new level for each unique y
    if (levelY.find(y) == levelY.end())
    {
      double cx = l.X();
      double cy = y;
      double cz = l.Z();

      double csx = ls.X();
      double csy = _size.Y() * levelSizeFactor;
      double csz = ls.Z();

      Level level;
      level.x = cx;
      level.y = cy;
      level.z = cz;
      level.sx = csx;
      level.sy = csy;
      level.sz = csz;
      level.buffer = _buffer;
      levelY[y] = level;
    }
    levelY[y].refs.push_back(name);
  }
  
  // print out plugin with all the levels
  std::stringstream out;
  out << "    <!-- Auto generated: level_generator "
      << _size << " " << _buffer << " -->\n";
  out << "    <plugin name=\"ignition::gazebo\" filename=\"dummy\">\n";
  out << levelsStr(levelX);
  out << levelsStr(levelY);
  out << "    </plugin>";
  std::cout << out.str() << std::endl;

  
  // print debug visuals
  if (debugLevels)
  { 
    printDebugVisuals(levelX);
    printDebugVisuals(levelY);
  }
}
```

subt_ign/src/level_generator.cc
iche033: fixed. 500847a9e4f22861f0c737b3386ae60364523fb8

subt_ign/src/level_generator.cc
iche033: fixed. 500847a9e4f22861f0c737b3386ae60364523fb8

subt_ign/src/level_generator.cc
iche033: updated doc. It should be at minimum the size of a tile and used to determine the final size of level. 500847a9e4f22861f0c737b3386ae60364523fb8

subt_ign/src/level_generator.cc
iche033: updated thanks. 500847a9e4f22861f0c737b3386ae60364523fb8

iche033:
oops just saw pull request #423. Let me merge the changes

iche033:
done changes merged a106f09 and updated PR description to reflect the new arguments

subt_ign/src/level_generator.cc
```diff
@@ -0,0 +1,313 @@
...
+  outFile << strCopy;
+  outFile.close();
+}
+
+//////////////////////////////////////////////////
+int main(int argc, char **argv)
+{
+  if (argc != 7)
+  {
+    std::cerr << "Usage: dot_generator <path_to_world_sdf_file> "
```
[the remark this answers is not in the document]
iche033: fixed. e64fc3afd4a915883aa5000f1fb3ceb1876b989e

PR #427, opened 2020-05-20
Fixes a segfault during exit of SubT simulation

azeey:
Making an ign-transport request from within a destructor context seems to cause a segfault. This replaces the request with an event.

To test, run the following launch file with `ign launch -v 4` and after a few seconds `ctrl-c`. Before this PR, this would cause a segfault.

```
<?xml version="1.0"?>
<ignition version='1.0'>
  <env>
    <name>IGN_GAZEBO_SYSTEM_PLUGIN_PATH</name>
    <value>$LD_LIBRARY_PATH</value>
  </env>

  <!-- Unlimited Ignition Transport send message buffer -->
  <env>
    <name>IGN_TRANSPORT_SNDHWM</name>
    <value>0</value>
  </env>

  <!-- Unlimited Ignition Transport receieve message buffer -->
  <env>
    <name>IGN_TRANSPORT_RCVHWM</name>
    <value>0</value>
  </env>

  <plugin name="ignition::launch::GazeboServer"
          filename="libignition-launch-gazebo.so">
    
    <world_file>empty.sdf</world_file>
    <run>true</run>
    <!-- The SubT challenge logic plugin -->
    <plugin entity_name="empty"
            entity_type="world"
            filename="libGameLogicPlugin.so"
            name="subt::GameLogicPlugin">
      <world_name>empty</world_name>
    </plugin>
  </plugin>
</ignition>
```

### DIFF
--- a/subt_ign/src/GameLogicPlugin.cc
+++ b/subt_ign/src/GameLogicPlugin.cc
@@ -37,6 +37,7 @@
 #include <ignition/gazebo/components/World.hh>
 #include <ignition/gazebo/Conversions.hh>
 #include <ignition/gazebo/EntityComponentManager.hh>
+#include <ignition/gazebo/Events.hh>
 #include <ignition/gazebo/Util.hh>
 
 #include <ignition/common/Console.hh>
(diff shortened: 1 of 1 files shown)


nkoenig:
Nice. this works for me.

PR #437, opened 2020-05-29
Feature/model pr 434 improvements

Samu24:
@acschang 

PR addresses the following points as mentioned [here](https://github.com/osrf/subt/pull/383#issuecomment-634971668)

- Gas sensor on model
- Adjust camera parameters
- Enable and adjust battery plugin
- Add noise to `robot_base_twist`

Some comments:

- **Camera Parameters**: Please note the distortion is set to zero because we are using distortion corrected intrinsics for the camera this is due to the fact that we are using an equidistant model for distortion and gazebo only supports the RadTan model.

- **Battery plugin**: We control our robot via JointTorqueCommand. This does not seem to put the battery into discharging mode. Is this intended?

- **Noise on `robot_base_twist`**: The linear velocity standard deviation for x and y are bit high to feed them unfiltered into the controller. While it does not create stability issues, the controller commands the robot to keep on walking even in case of a zero twist command due to the noise. I reduced them to the point where the robot stops walking in case of a zero twist command.

Thanks for your feedback.


nkoenig:
Here is the fix to Gazebo for the battery plugin: https://github.com/ignitionrobotics/ign-gazebo/pull/165. Just to confirm, you are using `JointForceCommand`, correct? I'm asking because your description mentions `JointTorqueCommand`.

Samu24:
Yes this is a mistake in my comment. We are using `JointForceCommand`

Here is the code snippet:
```
// Apply Desired Torque
int jointIterator = 0;
for (Entity joint : this->dataPtr->joints) {
  auto torque = _ecm.Component<components::JointForceCmd>(joint);
  if (torque == nullptr) {
    _ecm.CreateComponent(joint, components::JointForceCmd ({0.0}));
  } else {
    torque->Data()[0] = this->dataPtr->desiredJointTorques(jointIterator);
  }
    jointIterator++;
}
```

acschang:
As noted in your comments, the battery state does not change to discharging mode when the vehicle is commanded.

The other issues pointed out in the original PR are resolved.

Samu24:
@acschang I think the fix for the plugin is merged now. Do you need anything more from our side?

acschang:
No, thank you for completing these changes.  I'm going to merge your changes into your original PR now.

PR #440, opened 2020-06-02
Update to visibility RF model

acschang:
Updated `VisbilityRFModel.cc` to factor in radio characteristics alongside the visibility graph.

Before:
before_breadcrumb

After:
default_sw01_loc1


acschang:
> Looks pretty good to me. Nice to visualize a more accurate RF model. My only question is why just 5 bytes in `num_bytes` and not a more realistic value? Not sure which value it is best but probably higher than that.

I agree 5 bytes is quite low.  I was looking to duplicate unit tests results with a very small message for verification.  I had intended to make these parameters configurable but I'm not sure where the best place to put this configuration in would be.  As a first step, increasing message size (number of bytes) would be a good first step.  The size of a `geometry_msgs/TransformStamped` message, or approximately 100 bytes, would be an adequate fixed example for now.

> Actually, I have another suggestion. How about targeting this pull request towards `cave_feature_release2`? This way we don't risk losing these changes if we keep them in a separate branch. When merged, we can merge this changes into `breadcrumb_models_caguero`.

I'll adjust the target.

PR #442, opened 2020-06-02
Add NIOSH worlds

iche033:
Added NIOSH EX and SR tunnels. 

There are two configurations per tunnel: Config A and Config B. So altogether 4 worlds:
  * SR Config A
  * SR Config B
  * EX Config A
  * EX Config B

The difference between Config A and Config B is the location of artifacts. 

NIOSH SR
niosh_seg01

NIOSH EX
niosh_seg02

Example usage:

```
ign launch -v 4 cloudsim_sim.ign worldName:=niosh_sr_config_a robotName1:=X1 robotConfig1:=X1_SENSOR_CONFIG_1
```

The two worlds are created from these two models on fuel:
https://app.ignitionrobotics.org/OpenRobotics/fuel/models/NIOSH%20Segment%2001
https://app.ignitionrobotics.org/OpenRobotics/fuel/models/NIOSH%20Segment%2002

They use the old NIOSH staging area (one without tall fence) since it's the only one that fits the entrance opening of these two NIOSH tunnels:
https://app.ignitionrobotics.org/OpenRobotics/fuel/models/NIOSH%20Staging%20Area

Note that the NIOSH models are two large meshes as opposed to individual tiles so we won't be able to generate dot / dat files for these worlds. If we want dot / dat, we'll need to crop these two NIOSH models into smaller tiles. 
**Update** NIOSH SR now split into sections. Added dat and dot in 1542641
**Update** NIOSH EX now split into sections. Added dat and dot in 8aad7c4

~~Another issue with large meshes is that they cause delay with mouse presses as since ign-gazebo performs ray-triangle intersection checks on every click. So navigating the user camera in these two worlds can be challenging.~~. Resolved.


iche033:
All new NIOSH tiles have been uploaded onto fuel and added to the worlds. The dot and dat files have also been generated. This PR is ready for review again.

iche033:
**Update** Renamed NIOSH worlds from Segment 01 and Segment 02 to NIOSH SR and NIOSH EX worlds. Also added 20 artifacts to each world configuration, and updated description of this PR.

nkoenig:
The staging area on the `sr` worlds needs to move the next tunnel on the right.

iche033:
the SR world tiles have been shifted so that the starting area is on the right tunnel. Also regenerated the dat files. 717eddb

iche033:
New niosh meshes have been uploaded onto fuel. 

Tweaked a few artifact positions to  work with the new niosh meshes. Also updated dat files.

PR #445, opened 2020-06-03
Robotika Kloubak K2 - fixes (lidar, odometry, battery power_load, ...)

m3d:


acschang:
The issues with the LIDAR returns have been corrected but I still an issue with both front and rear odometry being published on the same topic.  The changes to the `power_load` are sufficient.

m3d:
@acschang thanks. If you have suggestion how should front & rear odometry be resolved, please let me know. Kloubak is really like two differentially driven robots connected together:
kloubak-k2-qual
(motors are in wheels)

acschang:
I'm going to go ahead and merge this to the original PR.  I'll address the issue with odometry publishing on the same topic for both front and rear differential drives.

I will post a follow-up question for control of the vehicle on the original PR.

PR #449, opened 2020-06-05
Robotika Kloubak K2 - update specification

m3d:
Here is updated Kloubak K2 specification with links to some data and videos from @tajgr

### DIFF
--- a/submitted_models/robotika_kloubak_sensor_config_1/specification.md
+++ b/submitted_models/robotika_kloubak_sensor_config_1/specification.md
@@ -24,16 +24,34 @@ The following specific sensors are declared payloads of this vehicle:
 * Wheel odometry - wheel encoders modelled by `pose-publisher` plugin.
 
 ## Control
-Kloubak is controlled by the open-source Osgar framework.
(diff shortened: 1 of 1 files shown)


acschang:
Thank you for the in-depth reply regarding the control the Kloubak vehicle and updating the `specifications.md` file with additional information.  Some additional items to address in the `specifications.md` file are:

- In `specifications.md`, it is noted that a difference is "It does not model angle sensor in the main joint and the second IMU is used instead" however the angle of joint can be extracted using a transform lookup from `<robot_name>/chassis_front` to `<robot_name>/chassis_rear`.  Please include this transform lookup as the simulated alternative to the angle sensor.
- Please note the joint limit between the front and rear chassis in `specifications.md`.  The maximum joint angle is approximately 70 degrees in the `model.sdf` for reference.  If there are any recommendations for standard operating procedure regarding maximum joint angle (i.e. +-45 degrees) to maintain control and accurate odometry (i.e. during 2WD mode) please note them.
- Add a link in `specifications.md` to an example Kloubak controller used by Robotika.  This can direct users to the OSGAR repository.

PR #453, opened 2020-06-09
Setting max_velocities/accelerations for X1 and X2

caguero:
Requires [this](https://github.com/ignitionrobotics/ign-gazebo/pull/161) pull request.

Velocity and accelerations limits for X1 and X2.


nkoenig:
> This looks good on `ign-gazebo-2.20.0`. Will these limits be expanded to other `*.ign` files and then to competitor submitted vehicles?

That's the idea.

nkoenig:
Added limits to other launch files.

PR #454, opened 2020-06-10
Add cave world generator

iche033:
**Update**: Updated code to work with both Cave Type A and Type B tiles. Added an extra `-t <type>` arg that lets user specify the type of cave to create: `anastomotic` - creates world using Type A tiles, `rectilinear` - creates world using Type B tiles, and `curvilinear` - creates world using Type A and Type B tiles.  Also updated PR description.

---

Added `world_generator_cave` that generates a world from type A and type B tiles. The tool constructs a world from a collection of prefab world sections and certain individual tiles. Note that the code is currently specific to Cave Type A and Type B Tiles. 

Some options are available:

```
Usage: world_generator_cave [options]
Options:
    -h		 Print this help message
    -o <file>	 Output sdf filename
    -s <seed>	 Seed
    -c <count>	 Min tile count
    -n <name>	 World name
    -t <type>	 Cave Type:
             	    'anastomotic' or 'a',
             	    'curvilinear' or 'c',
             	    'rectilinear' or 'r'
    -g		 Generate sdf with GUI plugin

```

You can generate a different world by setting a different seed value and minimum tile count.

Note the optional `g` arg that lets users generate a world sdf with GUI plugin, which is useful for viewing the world without launching the whole subt stack. Leave out the `g` arg when generating the final world.

Example usage:

```
./install/share/subt_ign/world_generator_cave -g -t a -c 60 -n cave_test -s 25 -o cave_test.sdf
```

If you do not have the Cave Type A and Type B tiles in your fuel cache, the tool will go and download them as it needs to compute bounding box info for these tiles in order to do intersection checks.

View the world using ign-gazebo

```
ign gazebo -v 4 cave_test.sdf
```

world_generator_type_a_25

### DIFF
--- a/subt_ign/CMakeLists.txt
+++ b/subt_ign/CMakeLists.txt
@@ -298,15 +298,15 @@ target_include_directories(level_generator
 install(TARGETS level_generator
   DESTINATION ${CATKIN_PACKAGE_SHARE_DESTINATION})
 
-add_executable(world_generator_type_a src/world_generator_type_a.cc src/ConnectionHelper.cc)
-target_link_libraries(world_generator_type_a
+add_executable(world_generator_cave src/world_generator_cave.cc src/ConnectionHelper.cc)
+target_link_libraries(world_generator_cave
   ignition-math6::ignition-math6
   ignition-common3::requested
   ignition-fuel_tools3::ignition-fuel_tools3
   sdformat8::sdformat8)
-target_include_directories(world_generator_type_a
+target_include_directories(world_generator_cave
   PRIVATE ${CATKIN_DEVEL_PREFIX}/include)
-install(TARGETS world_generator_type_a
+install(TARGETS world_generator_cave
   DESTINATION ${CATKIN_PACKAGE_SHARE_DESTINATION})
 
 
--- a/subt_ign/CMakeLists.txt
+++ b/subt_ign/CMakeLists.txt
@@ -27,7 +27,7 @@ set(IGN_GAZEBO_VER ${ignition-gazebo2_VERSION_MAJOR})
 file(MAKE_DIRECTORY ${CATKIN_DEVEL_PREFIX}/include)
 
 find_package(ignition-common3 REQUIRED COMPONENTS graphics)
-find_package(ignition-fuel_tools4 REQUIRED)
+find_package(ignition-fuel_tools3 REQUIRED)
 find_package(ignition-math6 REQUIRED)
 find_package(ignition-msgs4 REQUIRED)
 find_package(ignition-transport7 REQUIRED)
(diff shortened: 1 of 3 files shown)


subt_ign/CMakeLists.txt
```diff
@@ -297,6 +298,17 @@ target_include_directories(level_generator
 install(TARGETS level_generator
   DESTINATION ${CATKIN_PACKAGE_SHARE_DESTINATION})
 
+add_executable(world_generator_type_a src/world_generator_type_a.cc src/ConnectionHelper.cc)
+target_link_libraries(world_generator_type_a
+  ignition-math6::ignition-math6
+  ignition-common3::requested
```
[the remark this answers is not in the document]
mjcarroll: Ah, very cool! Today I learned.

subt_ign/CMakeLists.txt
```diff
@@ -26,7 +26,8 @@ set(IGN_GAZEBO_VER ${ignition-gazebo2_VERSION_MAJOR})
 
 file(MAKE_DIRECTORY ${CATKIN_DEVEL_PREFIX}/include)
 
-find_package(ignition-common3 REQUIRED)
+find_package(ignition-common3 REQUIRED COMPONENTS graphics)
+find_package(ignition-fuel_tools4 REQUIRED)
```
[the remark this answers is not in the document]
iche033: good catch, fixed in 3972783

angelacmaio (approved):

subt_ign/src/world_generator_cave.cc
```diff
@@ -0,0 +1,1539 @@
...
+      std::string name = "tile_" + std::to_string(++tileNo);
+
+      // convert tile pose to world coordinates
+      math::Pose3d pose = t.model.Pose() + s.pose;
+      std::string uri =
+          "https://fuel.ignitionrobotics.org/1.0/OpenRobotics/models/" +
+          t.tileType;
+
+      ss << "    <include>\n";
+      ss << "      <static>true</static>\n";
```
angelacmaio, line 1346: Should all tiles be static?

subt_ign/src/world_generator_cave.cc
```diff
@@ -0,0 +1,1539 @@
...
+    else if (uf.tileType == CAVE_TYPE_B)
+    {
+      uri = capUriTypeB;
+      // the cap origin is at an offset and so it needs to be placed at
+      // center of the next tile position instead of where the opening is
+      pose = math::Pose3d(pos + rot*math::Vector3d(12.5, 0, 0),
+          math::Quaterniond(0, 0, IGN_PI/2)*rot);
+    }
+    ss << "    <include>\n";
+    ss << "      <static>true</static>\n";
```
angelacmaio, line 1396: Should all caps be static?

subt_ign/src/world_generator_cave.cc
iche033: yes they should be static. The models are actually already marked as static in the model.sdf file so this is probably redundant. I added this to be consistent with the existing world files.

subt_ign/src/world_generator_cave.cc
iche033: yes same as the tiles

PR #465, opened 2020-06-15
Use github actions to build and push images to dockerhub

j-rivero:
The PR implements an auto-mechanism to generate the docker images and push them into the dockerhub. The way it currently works:

 * Requires credentials setup in subt repository (dockerhub user + token)
 * Trigger the build when someone push a tag with the name `dockerhub_release_*` (we can use any other interaction to trigger the build, tag seemed to me flexible enough but we could also consider the `releases` mechanism in github)
 * The job will trigger four different builds, one per docker image, define in `image_name` (see [my testing repo](https://github.com/j-rivero/subt/actions/runs/136258747))
 * It uses the docker/build.sh script
 * There is a special threatment for `subt_sim_entry` which seems to be the one producing the `latest` image (see how actions [are ignoring these steps for other images](https://github.com/j-rivero/subt/runs/767112433?check_suite_focus=true)).

Please be sure that [the images in my dockerhub ](https://hub.docker.com/r/jlrivero/subt/tags)matches the one in subt. 

TODO: I did not implement anything related to `-private` repository. Please let me know if we need to do it.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -28,7 +28,7 @@ jobs:
       - name: Tag image
         run: docker tag ${{ matrix.image_name }}:latest ${{ env.registry }}:${{ matrix.image_name }}_latest
 
-      # only reserverd for subt_sim_entry
+      # only reserved for subt_sim_entry
       - name: Tag latest image
         run: docker tag ${{ matrix.image_name }}:latest ${{ env.registry }}/subt:latest
         if: matrix.image_name == 'subt_sim_entry'
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -42,7 +42,6 @@ jobs:
       - name: Push to docker
         uses: actions-hub/docker@master
         env:
-          NAME: ${{ matrix.image_name }}
           TAG: ${{ env.registry }}:${{ matrix.image_name }}_latest
         with:
           args: push ${IMAGE_TAG}
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,4 +1,4 @@
-name: Relase to dockerhub
+name: Release to dockerhub
 on:
   push:
     tags:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,62 @@
+name: Relase to dockerhub
+on:
+  push:
+    tags:
+      - dockerhub_release_*  # Launch dockerhub push on dockerhub_release_* tags
+jobs:
+  build_docker_image:
+    runs-on: ubuntu-latest
+    env:
+      registry: osrf/subt-virtual-testbed
+    strategy:
+      matrix:
+        image_name: ['cloudsim_sim', 'cloudsim_bridge', 'subt_shell', 'subt_sim_entry']
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v2
+
+      - name: Install base dependencies
+        if: ${{ success() }}
+        run: |
+          sudo apt update;
+          sudo apt-get install -y docker
+
+      - name: Run build script
+        if: ${{ success() }}
+        run: |
+            cd docker
+            bash +xe ./build.bash ${{ matrix.image_name }}
+
+      - name: Tag image
+        if: ${{ success() }}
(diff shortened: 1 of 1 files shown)


j-rivero:
Together with @caguero we found that NAME variable was not really being used for anything in the docker push  and the checks for success are redundant since the default mode is to fail the build if one of the steps return failure.

.github/workflows/main.yml
```diff
@@ -0,0 +1,56 @@
+name: Release to dockerhub
+on:
+  push:
+    tags:
+      - dockerhub_release_*  # Launch dockerhub push on dockerhub_release_* tags
```
caguero, line 5: Nitpick: I wonder if we should use a regular tag here, as opposed to a dockerhub_release tag specific.

.github/workflows/main.yml
```diff
@@ -0,0 +1,56 @@
...
+      - name: Login to docker hub
+        uses: actions-hub/docker/login@master
+        env:
+          DOCKER_USERNAME: ${{ secrets.DOCKER_USERNAME }}
+          DOCKER_PASSWORD: ${{ secrets.DOCKER_PASSWORD }}
+
+      - name: Push to docker
+        uses: actions-hub/docker@master
+        env:
+          NAME: ${{ matrix.image_name }}
```
caguero, line 45: We convinced ourselves that this parameter isn't used, right?

.github/workflows/main.yml
j-rivero: whatever you think would be the most useful

.github/workflows/main.yml
j-rivero: ouch, forgot that one. thanks

.github/workflows/main.yml
caguero: How about any tag?

nkoenig: I'm not very familiar with github actions. Does this line mean that the action will only be triggered when a `dockerhub_release_` tag is created?

.github/workflows/main.yml
```diff
@@ -0,0 +1,55 @@
...
+    tags:
+      - dockerhub_release_*  # Launch dockerhub push on dockerhub_release_* tags
+jobs:
+  build_docker_image:
+    runs-on: ubuntu-latest
+    env:
+      registry: osrf/subt-virtual-testbed
+    strategy:
+      matrix:
+        image_name: ['cloudsim_sim', 'cloudsim_bridge', 'subt_shell', 'subt_sim_entry']
```
nkoenig, line 13: You can remove `subt_shell`.

.github/workflows/main.yml
```diff
@@ -0,0 +1,55 @@
...
+        run: |
+            cd docker
+            bash +xe ./build.bash ${{ matrix.image_name }}
+
+      - name: Tag image
+        run: docker tag ${{ matrix.image_name }}:latest ${{ env.registry }}:${{ matrix.image_name }}_latest
+
+      # only reserved for subt_sim_entry
+      - name: Tag latest image
+        run: docker tag ${{ matrix.image_name }}:latest ${{ env.registry }}/subt:latest
```
nkoenig, line 33: Should `${{ env.registry }}/subt:latest` be `${{ env.registry }}:latest`?

.github/workflows/main.yml
caguero: That's my understanding.

.github/workflows/main.yml
```diff
@@ -0,0 +1,55 @@
...
+      matrix:
+        image_name: ['cloudsim_sim', 'cloudsim_bridge', 'subt_shell', 'subt_sim_entry']
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v2
+
+      - name: Install base dependencies
+        run: |
+          sudo apt update;
+          sudo apt-get install -y docker
```
caguero, line 21: I recommend to check that this line really installs `docker`. This used to work for me on VRX on `ubuntu-latest` but that's no longer the case (maybe something changed in the Ubuntu package?). I installed it using this action:

```
- name: Install Docker
        uses: docker-practice/actions-setup-docker@master
        with:
          docker_version: 18.09
          docker_channel: stable
```

nkoenig:
Thanks for all the work on this. I'm going to close this PR since the docker build process has changed, and I really need to do local testing of docker images before pushing them to dockerhub.

PR #471, opened 2020-06-22
Update Marble HD2 gimbal meshes and sensor pose

iche033:
* Updated to use the photogrammetry meshes for the gimbal system.
* Updated the pose of camera and lidar sensors to align with the new meshes
* Added some friction to gimbal pan joint as it was drifting slightly on its own on launch

Screenshot:

marble_hd2_meshes

One thing I'm not sure about is where the planar lidar should be as I don't see the device in the new mesh. So I placed it roughly around where it was in the original model, see image below: the planar lidar is represented using a red box and the rviz visualization on the right shows some self-intersection which may already be present (but slightly different) in the original version.

marble_hd2_planar_lidar

### DIFF
--- a/submitted_models/marble_hd2_sensor_config_1/meshes/hd2.dae
+++ b/submitted_models/marble_hd2_sensor_config_1/meshes/hd2.dae
@@ -12,7 +12,7 @@
     <revision/>
     <subject/>
     <title/>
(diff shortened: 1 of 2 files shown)


acschang:
The adjusted RGBD camera locations are accurate and resolve the issue.

PR #473, opened 2020-06-23
Update Marble And Kloubak models.

nkoenig:
This updates two `specifications.md` files and added missing optical frames to the QAV500 model.


iche033:
I updated `vehicle_topics.ign` to also bridge the tof `camera_info` topics so that depth cloud can be visualized in rviz. 4f914df

PR #479, opened 2020-06-25
Start of an event logger

nkoenig:
This will log information whenever a robot enters or leaves a performer detectors. Example output:

```
- event:
    time_sec: 36
    detector: performer_detector_staging
    robot: X1
    state: enter
    pos:
        x: -2.99869
        y: 2.67071
        z: 0.137346
- event:
    time_sec: 45
    detector: performer_detector_staging
    robot: X1
    state: exit
    pos:
        x: 3.00125
        y: 0.842902
        z: 0.183054
```

### DIFF
--- a/subt_ign/src/GameLogicPlugin.cc
+++ b/subt_ign/src/GameLogicPlugin.cc
@@ -153,6 +153,10 @@ class subt::GameLogicPluginPrivate
   /// \return The time point used to calculate the elapsed real time.
   public: std::chrono::steady_clock::time_point UpdateScoreFiles() const;
 
+  /// \brief Performer detector subscription callback.
+  /// \param[in] _msg Pose message of the event.
+  public: void OnEvent(const ignition::msgs::Pose &_msg);
+
   private: bool PoseFromArtifactHelper(const std::string &_robot,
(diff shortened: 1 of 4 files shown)


subt_ign/src/GameLogicPlugin.cc
```diff
@@ -414,12 +426,46 @@ void GameLogicPlugin::Configure(const ignition::gazebo::Entity & /*_entity*/,
...
+    }
+  }
+
+  this->eventStream
+    << "- event:\n"
+    << "    time_sec: " << _msg.header().stamp().sec() << "\n"
+    << "    detector: " << frameId << "\n"
+    << "    robot: " << _msg.name() << "\n"
+    << "    state: " << state << "\n"
+    << "    pos:\n"
```
azeey, line 463: The position published by the performer_detector system is relative to the pose of the model containing the system. Will that be an issue?

nkoenig: Ahh, thanks. I was printing it because the information was there. It's not important to have that bit. Removed in: 7888f4fa1995f01c49c51816be4b95b11c3811cf

PR #450, opened 2020-06-08
Model Submission for R2 configuration 1 from Explorer

HongbiaoZ:

PR #482, opened 2020-07-09
Submitted models/explorer r2 sensor config1

nkoenig:
This will replace pull request #450.


nkoenig:
@acschang , this is mostly ready for review. We are still working out the diff drive system.

azeey:
I've made changes to the wheel parameters to improve steering https://github.com/osrf/subt/pull/482/commits/c24ad9428124f3e1fa5a29c88ec61eefbdc3158c. Namely

- Change slip1 to 0.046 to match X1 vehicles, but scaled by the mass of this vehicle. The calculation is `slip_R2 = mass_R2/mass_X1 * slip_X1`.
- Remove velocity and effort limits because those are handled by the DiffDrive system

HongbiaoZ:
I have changed rotation angle limit between Center Pivot and base_link to -30-3- degree. Modify the mass of base_link and Center Pivot to match the value of physical robot. I changed the "y" value of the poses for each wheel from 0.343 to 0.363. Because of some slight errors on our model, the wheel are too close to the robot body so that it's really hard for it to turn around. Now it should be able to turn around smoothly.

nkoenig:
This looks good to me.

iche033:
I bridged `depth/camera_info` topic in bb7101c so DepthClouds can be visualized in rviz

PR #484, opened 2020-07-15
Log more data from each run

iche033:
More data are now logged to `run.yml` in `/tmp/ign/logs`. The new `run.yml` log does contain some duplicate data from `score.yml` and `summary.yml` but these existing two log files remain untouched so the changes here do not break existing post processing scripts that use them. 

The following data are logged:

1. Number of artifacts found.
2. Robot count
3. Unique robot count (for example, a team of two X1_SENSOR_CONFIG_1 robots would have a unique count of 1).
4. Total simulation time.
5. Total real time.
6. Number of artifact report attempts
7. Number of duplicate artifact reports.
8. Closest artifact report:
     a. Distance in meters.
     b. True position of the artifact.
     c. Reported position.
     d. Artifact type and name.
9. First artifact report time.
10. Last artifact report time.
11. Mean time between success reports

To manually test this, you can use the `test_score` test executable, see [README.md](https://github.com/osrf/subt/tree/master/subt_ign/test). The test itself may be flaky but the `run.yml` log file should still be correctly generated.

### DIFF
--- a/subt_ign/src/GameLogicPlugin.cc
+++ b/subt_ign/src/GameLogicPlugin.cc
@@ -241,16 +241,13 @@ class subt::GameLogicPluginPrivate
   /// reported pos, distance between true pos and reported pos
   public: std::tuple<std::string, std::string, ignition::math::Vector3d,
       ignition::math::Vector3d, double> closestReport =
-    {"", "", ignition::math::Vector3d(), ignition::math::Vector3d(),
-    std::numeric_limits<double>::infinity()};
+    {"", "", ignition::math::Vector3d(), ignition::math::Vector3d(), -1};
 
   /// \brief First artifact report time
-  public: double firstReportTime =
-      std::numeric_limits<double>::infinity();
+  public: double firstReportTime = -1;
 
   /// \brief Last artifact report time
-  public: double lastReportTime =
-      std::numeric_limits<double>::infinity();
+  public: double lastReportTime = -1;
 
   /// \brief A mutex.
   public: std::mutex mutex;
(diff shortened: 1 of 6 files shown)


subt_ign/src/GameLogicPlugin.cc
```diff
@@ -1092,6 +1150,77 @@ GameLogicPluginPrivate::UpdateScoreFiles() const
...
+      << YAML::Value << std::get<1>(this->closestReport)
+      << YAML::Key << "true_pos"
+      << YAML::Value << artifactPos.str()
+      << YAML::Key << "reported_pos"
+      << YAML::Value << reportedPos.str()
+      << YAML::Key << "distance"
+      << YAML::Value << std::get<4>(this->closestReport)
+      << YAML::EndMap;
+  out << YAML::Key << "first_artifact_report";
+  out << YAML::Value << this->firstReportTime;
```
nkoenig, line 1206: Can you make `firstReportTime` and `lastReportTime` output `0` instead of `inf` in the situation when no artifacts are reported?

subt_ign/src/GameLogicPlugin.cc
```diff
@@ -1092,6 +1150,77 @@ GameLogicPluginPrivate::UpdateScoreFiles() const
...
+      << YAML::Key << "name"
+      << YAML::Value << std::get<0>(this->closestReport)
+      << YAML::Key << "type"
+      << YAML::Value << std::get<1>(this->closestReport)
+      << YAML::Key << "true_pos"
+      << YAML::Value << artifactPos.str()
+      << YAML::Key << "reported_pos"
+      << YAML::Value << reportedPos.str()
+      << YAML::Key << "distance"
+      << YAML::Value << std::get<4>(this->closestReport)
```
nkoenig, line 1203: Output a value of `-1` instead of `inf` when no artifacts have been reported.

subt_ign/src/GameLogicPlugin.cc
iche033: done. c65170d

subt_ign/src/GameLogicPlugin.cc
iche033: done. c65170d

PR #495, opened 2020-07-23
Improve thread safety when logging

nkoenig:
This just makes sure that the event stream is correct in the case of multiple simultaneous events.

### DIFF
--- a/subt_ign/src/GameLogicPlugin.cc
+++ b/subt_ign/src/GameLogicPlugin.cc
@@ -161,6 +161,10 @@ class subt::GameLogicPluginPrivate
   /// \param[in] _msg Pose message of the event.
   public: void OnEvent(const ignition::msgs::Pose &_msg);
 
+  /// \brief Log an event to the eventStream.
+  /// \param[in] _event The event to log.
+  public: void LogEvent(const std::string &_event);
+
   /// \brief Marsupial detach subscription callback.
   /// \param[in] _msg Detach message.
   /// \param[in] _info Message information.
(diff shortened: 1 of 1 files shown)


subt_ign/src/GameLogicPlugin.cc
```diff
@@ -1189,13 +1214,15 @@ void GameLogicPluginPrivate::Finish()
     this->logStream.flush();
     this->eventStream.flush();
```
iche033, line 1215: I think this flush call can be moved to `LogEvent` function

subt_ign/src/GameLogicPlugin.cc
```diff
@@ -1375,6 +1402,13 @@ bool GameLogicPluginPrivate::OnPoseFromArtifact(
   return result;
 }
 
+/////////////////////////////////////////////////
+void GameLogicPluginPrivate::LogEvent(const std::string &_event)
+{
+  std::lock_guard<std::mutex> lock(this->eventMutex);
+  this->eventStream << _event;
```
iche033, line 1409: add `this->eventStream.flush();` here?

subt_ign/src/GameLogicPlugin.cc
nkoenig: Removed: 762e1dd713a9e608ed1e4b302deb8273df15fc1d

subt_ign/src/GameLogicPlugin.cc
nkoenig: Added: 762e1dd713a9e608ed1e4b302deb8273df15fc1d

PR #507, opened 2020-07-28
Adds a competition countdown clock

nkoenig:
Introduces a `/subt/clock` topic on both Ignition and ROS.

The Ignition message look like:

```
header {
  data {
    key: "phase"
    value: "run"
  }
}
sim {
  sec: 5
}
```

The ROS message looks like:

```
phase: "run"
data: 
  secs: 5
  nsecs:         0
```

The phases are: `warmup`, `run`, `finished`. 

The clock counts down the time left in a phase.

## To test

1. `ign launch cave_circuit.ign worldName:=simple_cave_01 robotName1:=X1 robotConfig1:=X1_SENSOR_CONFIG_1 durationSec:=10`

2. `ign topic -et /subt/clock`

3. `rostopic echo /subt/clock`

4. Teleop the robot into the tunnel using: `roslaunch subt_example teleop.launch`

5. The duration is set to 10seconds, so you should see all three phases.

### DIFF
--- a/subt_ign/src/GameLogicPlugin.cc
+++ b/subt_ign/src/GameLogicPlugin.cc
@@ -281,9 +281,6 @@ class subt::GameLogicPluginPrivate
   /// \brief Ignition transport competition clock publisher.
   public: transport::Node::Publisher competitionClockPub;
 
-  /// \brief Ignition transport warmup clock publisher.
-  public: transport::Node::Publisher warmupClockPub;
-
   /// \brief Logpath.
   public: std::string logPath{"/dev/null"};
 
@@ -465,11 +462,7 @@ void GameLogicPlugin::Configure(const ignition::gazebo::Entity & /*_entity*/,
     this->dataPtr->node.Advertise<ignition::msgs::StringMsg>("/subt/start");
 
   this->dataPtr->competitionClockPub =
-    this->dataPtr->node.Advertise<ignition::msgs::Time>(
-        "/subt/clock/competition");
-
-  this->dataPtr->warmupClockPub =
-    this->dataPtr->node.Advertise<ignition::msgs::Time>("/subt/clock/warmup");
+    this->dataPtr->node.Advertise<ignition::msgs::Clock>("/subt/clock");
 
   this->dataPtr->publishThread.reset(new std::thread(
         &GameLogicPluginPrivate::PublishScore, this->dataPtr.get()));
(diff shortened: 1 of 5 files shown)


subt_ign/src/GameLogicPlugin.cc
```diff
@@ -696,6 +713,25 @@ void GameLogicPlugin::PostUpdate(
...
+      competitionClockMsg.mutable_header()->add_data();
+    mapData->set_key("phase");
+    if (this->dataPtr->started)
+    {
+      mapData->add_value(this->dataPtr->finished ? "finished" : "run");
+      auto secondsRemaining = std::chrono::duration_cast<std::chrono::seconds>(
+          remainingCompetitionTime);
+      competitionClockMsg.mutable_sim()->set_sec(secondsRemaining.count());
+    }
+    else if (!this->dataPtr->finished)
```
azeey, line 727: Do we have a case where `started = false` and `finished = true`?

subt_ros/package.xml
```diff
@@ -24,6 +24,7 @@
   <exec_depend>compressed_image_transport</exec_depend>
   <exec_depend>message_runtime</exec_depend>
   <exec_depend>topic_tools</exec_depend>
+  <exec_depend>std_msgs</exec_depend>
```
azeey, line 27: This should probably go in `<depend>`.

subt_ign/src/GameLogicPlugin.cc
```diff
@@ -458,6 +461,9 @@ void GameLogicPlugin::Configure(const ignition::gazebo::Entity & /*_entity*/,
   this->dataPtr->startPub =
     this->dataPtr->node.Advertise<ignition::msgs::StringMsg>("/subt/start");
 
+  this->dataPtr->competitionClockPub =
+    this->dataPtr->node.Advertise<ignition::msgs::Clock>("/subt/clock");
```
azeey, line 465: Consider changing the topic name to indicate that this is a count down clock so as to avoid confusion with `/clock`.

subt_ros/package.xml
nkoenig: Without this, catkin gives the following error:

```
CMake Error at /opt/ros/melodic/share/catkin/cmake/catkin_package.cmake:224 (message):
  catkin_package() DEPENDS on the catkin package 'std_msgs' which must
  therefore be listed as a run dependency in the package.xml
Call Stack (most recent call first):
  /opt/ros/melodic/share/catkin/cmake/catkin_package.cmake:102 (_catkin_package)
  subt/subt_ros/CMakeLists.txt:33 (catkin_package)
```

subt_ign/src/GameLogicPlugin.cc
nkoenig: Looks like a team could trigger finish before starting. Handled that case in 607c6de91d9baa5483bb824e18ba42aae172b934

subt_ign/src/GameLogicPlugin.cc
nkoenig: Changed to `run_clock` 607c6de91d9baa5483bb824e18ba42aae172b934

PR #519, opened 2020-08-02
Extended GameLogicPlugin to check for vehicle flips

adlarkin:
I'm checking for vehicle flips by checking the angle between the world's z-axis and the vehicle's z-axis (if the angle is close to 180 degrees, then the robot is probably flipped).

### DIFF
--- a/subt_ign/src/GameLogicPlugin.cc
+++ b/subt_ign/src/GameLogicPlugin.cc
@@ -24,6 +24,7 @@
 
 #include <algorithm>
 #include <chrono>
+#include <cmath>
 #include <map>
 #include <mutex>
 #include <utility>
@@ -48,6 +49,8 @@
 #include <ignition/common/Util.hh>
 #include <ignition/common/Time.hh>
 #include <ignition/math/Pose3.hh>
+#include <ignition/math/Quaternion.hh>
+#include <ignition/math/Vector3.hh>
 #include <ignition/transport/Node.hh>
 #include <sdf/sdf.hh>
 
@@ -1879,16 +1882,18 @@ std::ofstream &GameLogicPluginPrivate::Log()
 void GameLogicPluginPrivate::CheckRobotFlip(const std::string &_name,
                                             const ignition::math::Pose3d &_pose)
 {
-  // radian equivalents
-  static const double k170_deg = 2.96706;
-  static const double k190_deg = 3.316126;
+  // get the vehicle's z-axis in the world frame
+  ignition::math::Vector3d a = _pose.Rot() * ignition::math::Vector3d(0,0,1);
 
-  auto orientation = _pose.Rot().Euler();
-  auto roll = std::abs(orientation.X());
-  auto pitch = std::abs(orientation.Y());
+  // use dot product to get the angle between
+  // the vehicle's z-axis and the world's z-axis
+  // (I believe this simplifies to cos_theta = a.Z(),
+  // which makes sense since we only want the z component)
+  auto cos_theta = (a.X() * 0) + (a.Y() * 0) + (a.Z() * 1);
 
-  if ((roll >= k170_deg && roll <= k190_deg) ||
-      (pitch >= k170_deg && pitch <= k190_deg))
+  // cos_theta == -1 implies that the robot/world z-axis are in opposite directions,
+  // which probably means that the robot has flipped
+  if (std::abs(-1 - cos_theta) <= 0.1)
   {
     std::ostringstream stream;
     stream
--- a/subt_ign/src/GameLogicPlugin.cc
+++ b/subt_ign/src/GameLogicPlugin.cc
@@ -297,6 +297,11 @@ class subt::GameLogicPluginPrivate
   /// \brief A map of robot name and its previous pose
   public: std::map<std::string, ignition::math::Pose3d> robotPrevPose;
 
+  /// \brief A map of robot name to its flip information.
+  /// The value is a pair stating the sim time where the most recent flip started,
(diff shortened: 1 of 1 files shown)


subt_ign/src/GameLogicPlugin.cc
```diff
@@ -1865,3 +1874,28 @@ std::ofstream &GameLogicPluginPrivate::Log()
...
+/////////////////////////////////////////////////
+void GameLogicPluginPrivate::CheckRobotFlip(const std::string &_name,
+                                            const ignition::math::Pose3d &_pose)
+{
+  // radian equivalents
+  static const double k170_deg = 2.96706;
+  static const double k190_deg = 3.316126;
+
+  auto orientation = _pose.Rot().Euler();
+  auto roll = std::abs(orientation.X());
```
azeey, line 1887: I think Euler angles can be unreliable here because, for example, `(pi, pi, 0) == (0, 0, pi)`, but the vehicle in this case is not flipped. My suggestion would be to multiply the rotation with the unit `z` vector and then take the dot product of the result with the unit `z` vector. The result is the cosine of the angle between the world `+z` axis and the vehicle's `z` axis. If the vehicle is flipped, the value should be close to -1.0.

adlarkin: Thanks for pointing that out; I was not aware of that issue with Euler angles.

> My suggestion would be to multiply the rotation with the unit z vector and then take the dot product of the result with the unit z vector. The result is the cosine of the angle between the world +z axis and the vehicle's z axis.

If I am understanding correctly, it sounds like you're suggesting that I use this dot product equation (for two vectors **a** and **b**), isolating `cos(theta)`:

**a** dot **b** = ||**a**|| ||**b**|| cos(theta)   --> cos(theta) = (**a** dot **b**) / (||**a**|| ||**b**||)

So, in this case, **a** would be the rotation, and **b** would be the unit z vector (0,0,1). Am I understanding this correctly?

azeey: Yeah, the second part of what I was saying would use the dot product equation, but the first part would use the rotation matrix (or quaternion) to first rotate the vehicle's z axis (the unit z vector in the vehicle's local frame) into the world frame. You can use `_pose.Rot()` as a rotation matrix to multiply the unit z vector. So if **R** = `_pose.Rot()` and  **z** is the unit z vector (`0, 0, 1`), then
**a** = **R** * **z**
Now **a** is the vehicle's z axis in the world frame. We can then take the dot product of that with the unit z vector to find if the two are parallel:
cos(theta) = **a** dot **z**. These are all unit vectors so we can ignore magnitudes.

adlarkin: That makes sense; thanks for the detailed explanation!

I believe that I have just implemented what you described in [af0b090](https://github.com/osrf/subt/pull/519/commits/af0b0901532ef492eb433d3765f8c569056697ff). Would you mind taking a look at that commit and letting me know if I am still missing something? If it seems correct to you, I can make one last commit that cleans things up, and then I will re-request a review.

subt_ign/src/GameLogicPlugin.cc
```diff
@@ -1865,3 +1877,30 @@ std::ofstream &GameLogicPluginPrivate::Log()
...
+                                            const ignition::math::Pose3d &_pose)
+{
+  // get the vehicle's z-axis in the world frame
+  ignition::math::Vector3d a = _pose.Rot() * ignition::math::Vector3d(0,0,1);
+
+  // use dot product to get the angle between
+  // the vehicle's z-axis and the world's z-axis
+  // (I believe this simplifies to cos_theta = a.Z(),
+  // which makes sense since we only want the z component)
+  auto cos_theta = (a.X() * 0) + (a.Y() * 0) + (a.Z() * 1);
```
azeey, line 1892: Yup, this looks good. And yes, it simplifies to `a.Z()`.

subt_ign/src/GameLogicPlugin.cc
```diff
@@ -777,6 +786,9 @@ void GameLogicPlugin::PostUpdate(
               std::chrono::duration_cast<std::chrono::steady_clock::duration>
               (std::chrono::seconds(s) + std::chrono::nanoseconds(ns));
 
+          // check if the robot has flipped
+          this->dataPtr->CheckRobotFlip(name, pose);
```
azeey, line 790: Should we store whether the robot is already flipped? Otherwise, once it flips, the log file will be flooded with flip events. It might also be a good idea to record an event only after the robot stays flipped for a duration of time to avoid false positives.

azeey, line 790: For one `PostUpdate`, this runs once for every sensor in a robot. Can we limit it to run just once for a robot?

adlarkin: I went ahead and moved the `CheckRobotFlip` call outside of the sensor portion in [1256865](https://github.com/osrf/subt/pull/519/commits/1256865b11f89d3a8393c5130f50261eaa93046a), so I believe it should only be called once per `PostUpdate` now.

adlarkin: I added a data structure (map) that keeps track of the robot's flip information in [40f1186](https://github.com/osrf/subt/pull/519/commits/40f1186627e45ad30f27ed98f358de4c92d17b67). I used this map to make sure a robot has been flipped for a few seconds before logging a flip event, and also used it to only log a flip event once per new flip.

nkoenig: This works for me.

PR #524, opened 2020-08-05
Fix warning produced by CI in CommsBrokerPlugin

j-rivero:
Together with https://github.com/ignition-tooling/release-tools/pull/257 (already enabled today) should produce a green CI

### DIFF
--- a/subt_ign/src/CommsBrokerPlugin.cc
+++ b/subt_ign/src/CommsBrokerPlugin.cc
@@ -324,7 +324,7 @@ void CommsBrokerPlugin::UpdateIfNewBreadcrumbs()
   if (newBreadcrumbFound)
   {
     std::set<ignition::math::Vector3d> breadcrumbPoses;
-    #pragma GCC diagnostic Push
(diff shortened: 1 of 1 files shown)


subt_ign/src/CommsBrokerPlugin.cc
```diff
@@ -324,8 +324,11 @@ void CommsBrokerPlugin::UpdateIfNewBreadcrumbs()
   if (newBreadcrumbFound)
   {
     std::set<ignition::math::Vector3d> breadcrumbPoses;
+    #pragma GCC diagnostic push
+    #pragma GCC diagnostic ignored "-Wunused-variable"
```
mjcarroll, line 328: If we aren't using the variable, I would prefer that we update the code, rather than suppressing the warning.

We should be able to do 

```
for (const auto& it: this->breadcrumbs)
{
  breadcrumbPoses.insert(it->second.Pos());
}
```

j-rivero: thanks! changed.

PR #531, opened 2020-08-06
Record marsupial pairs

nkoenig:
Record marsupial pairs in the summary.yaml file.

### DIFF
--- a/subt_ign/src/GameLogicPlugin.cc
+++ b/subt_ign/src/GameLogicPlugin.cc
@@ -28,6 +28,7 @@
 #include <mutex>
 #include <utility>
 
+#include <ignition/gazebo/components/DetachableJoint.hh>
 #include <ignition/gazebo/components/Model.hh>
 #include <ignition/gazebo/components/Name.hh>
 #include <ignition/gazebo/components/DepthCamera.hh>
@@ -416,6 +417,9 @@ class subt::GameLogicPluginPrivate
 
   /// \brief Event manager for pausing simulation
   public: EventManager *eventManager;
+
+  /// \brief The set of marsupial pairs.
+  public: std::map<std::string, std::string> marsupialPairs;
 };
 
 //////////////////////////////////////////////////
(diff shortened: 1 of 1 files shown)


subt_ign/src/GameLogicPlugin.cc
```diff
@@ -640,6 +644,31 @@ void GameLogicPlugin::PostUpdate(
...
+
+          auto parentName = _ecm.Component<gazebo::components::Name>(
+              parentModel->Data());
+
+          auto childName = _ecm.Component<gazebo::components::Name>(
+              childModel->Data());
+          // Store the marsupial pairs, and handle the special case of
+          // "X1_platform"
+          if (parentName->Data() != "X1_platform" &&
+              childName->Data() != "X1_platform")
```
azeey, line 664: The platform is actually going to be named `<name>_platform`, so maybe check if the name ends with `_platform`? If we filter out non-performers, we won't have to do this check.

subt_ign/src/GameLogicPlugin.cc
```diff
@@ -640,6 +644,31 @@ void GameLogicPlugin::PostUpdate(
     std::map<std::string, ignition::math::Pose3d>::iterator baseIter =
       this->dataPtr->poses.find(subt::kBaseStationName);
 
+    _ecm.Each<gazebo::components::DetachableJoint>(
```
azeey, line 647: We use detachable joints for making breadcrumbs and dynamic obstacles static after a period of time. We can maybe filter them out by checking if both the parent and child models are also performers.

subt_ign/src/GameLogicPlugin.cc
nkoenig: I think this fixes it: 9d1b038f198dab3f825d7780e6503e657076ab51

subt_ign/src/GameLogicPlugin.cc
nkoenig: I think this fixes it: 9d1b038f198dab3f825d7780e6503e657076ab51

subt_ign/src/GameLogicPlugin.cc
```diff
@@ -640,6 +645,52 @@ void GameLogicPlugin::PostUpdate(
...
+          bool childPerformer = false;
+          for (const auto &t :
+              _ecm.ChildrenByComponents(childModel->Data(),
+                gazebo::components::Performer()))
+          {
+            childPerformer = true;
+            break;
+          }
+
+          auto parentName = _ecm.Component<gazebo::components::Name>(
```
azeey, line 675: nit: `parentName` and `childName` are defined again below.

subt_ign/src/GameLogicPlugin.cc
```diff
@@ -1561,6 +1612,9 @@ std::chrono::steady_clock::time_point GameLogicPluginPrivate::UpdateScoreFiles()
 
   // Output a run summary
   std::ofstream summary(this->logPath + "/summary.yml", std::ios::out);
+  summary << "marsupials:\n";
```
azeey, line 1615: If there are no marsupials, this will just print `marsupials:`. Is that okay?

nkoenig: I think that is still valid yaml, and would indicate no marsupials.

subt_ign/src/GameLogicPlugin.cc
nkoenig: Fixed.

PR #537, opened 2020-08-07
Absolem effort limit

nkoenig:
Apply effort limits to the flippers.


angelacmaio:
@peci1 For your awareness, the effort limits are changed here to match the statement in `specifications.md` that the robot can support itself with the flippers, but cannot use them to lift itself on all four flippers.

peci1:
@nkoenig Have you found another solution to publish the static transforms removed in 692c? Without them, the URDF model would not work.

Also, joints `left_track_j` and `right_track_j` are so far always at zero position, but I plan to add support for the lockable differential which would allow them to move. I'd rather keep these values in the joint states...

PR #543, opened 2020-08-11
Absolem: Fix static transforms and joint states

peci1:
Commit 1ae93583d2ba84c40b1e8589dd647460a9e981b4 mistakenly removed a `robot_state_publisher` that handled publication of static transforms. This PR returns it back.

It also returns back joints `left_track_j` and `right_track_j` into the published joints states, since these joints are expected to be moving in the future.


peci1:
This is how RViz displays the robot with the current master branch:

file:///home/peci1/Obr%C3%A1zky/Sn%C3%ADmek%20obrazovky%20po%C5%99%C3%ADzen%C3%BD%202020-08-11%2003-35-49.png

acschang:
These new features will not be incorporated into the SubT Virtual Testbed prior to Cave Circuit. We will keep this PR open for consideration after Cave Circuit.

peci1:
At least the part of this PR that fixes the static transforms is not a new feature but a bug fix. It is not only about missing links in rviz, but alsou about the inability to correctly transform e.g. the realsense cloud to base_link.

acschang:
> At least the part of this PR that fixes the static transforms is not a new feature but a bug fix. It is not only about missing links in rviz, but alsou about the inability to correctly transform e.g. the realsense cloud to base_link.

The simulated realsense cloud is published in the `alpha/base_link/camera` frame which there exists a transform to `alpha/base_link`.

peci1:
Ah, you're right, the camera link is published by Gazebo relative to base_link. But for example our algorithms make use of the `laser_base` frame, which is a convenience frame that denotes the place where the lidar gimbal connects to the robot body. This frame is missing without this fix.

nkoenig:
The bounding box for this model is
```
Min[-0.608707 -0.296315 -0.188594] Max[0.608707 0.296317 0.615689]
```

peci1:
I updated this PR with master and it's ready for review.

peci1:
Another merge with master (no conflict).

iche033:
The changes look fine and I see the new frames published after adding `robot_state_publisher`.

 Just want to check if preserving fixed joints when converting urdf to sdf is an option? That'll let ignition's pose publisher system publish the frames but will likely make the model.sdf diff larger

```
<gazebo reference="base_to_link">
  <disableFixedJointLumping>true</disableFixedJointLumping> 
  <preserveFixedJoint>true</preserveFixedJoint>
</gazebo>
```

peci1:
>  if preserving fixed joints when converting urdf to sdf is an option

Yes, that would also work. I've discovered that while preparing our new model and am using it there. What about merging this PR now and I would utilize preserving of fixed joints in a next batch of absolem updates?

peci1:
I updated this PR to resolve conflicts with master. We're still interesting in merging this in. The robot_state_publisher approach has been approved in #867.

peci1:
The robot_state_publisher was returned back in #859 and we can live without the track joints in joint states, so I'm closing this PR.

PR #545, opened 2020-08-11
Path tracer

nkoenig:
Added a path tracer program that visualizes robot paths, artifacts, and artifact reports.

# Usage

1. You need to use Ignition Citadel. An updated docker image is in the works.

2. Run the SubT world using the `path_tracer.ign` launch file along with an IGN_PARTITION name of PATH_TRACER. For example:

```
 IGN_PARTITION=PATH_TRACER ign launch -v 4 path_tracer.ign worldName:=cave_qual
```
3.  Run this program by passing in the directory that contains the simulation log files. Optionally specify the number of milliseconds to sleep between time step playback. If not specified a value of 20ms will be used. For example, from within a subt catkin workspace:

```
./install/bin/path_tracer ~/Downloads/my_logs 30
```

Signed-off-by: Nate Koenig <nate@openrobotics.org>

PR #546, opened 2020-08-11
Fix event stream.

nkoenig:
Fix an error in the event stream, and also improve the artifact report events

The matches the event stream modification in pull request #545.

### DIFF
--- a/subt_ign/include/subt_ign/CommonTypes.hh
+++ b/subt_ign/include/subt_ign/CommonTypes.hh
@@ -66,8 +66,8 @@ namespace subt
   enum class ArtifactType : uint32_t
   {
     TYPE_BACKPACK = 0,
-    TYPE_DUCT,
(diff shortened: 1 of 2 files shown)


subt_ign/include/subt_ign/CommonTypes.hh
```diff
@@ -66,8 +66,8 @@ namespace subt
   enum class ArtifactType : uint32_t
   {
     TYPE_BACKPACK = 0,
-    TYPE_DUCT,
     TYPE_DRILL,
+    TYPE_DUCT,
```
angelacmaio, line 70: Was this change intentional?

nkoenig: Yes, so that the indices match this: https://github.com/osrf/subt/blob/master/subt_ign/src/GameLogicPlugin.cc#L78

PR #547, opened 2020-08-13
Change unique_robot count to be based on robot platform types

nkoenig:
Log unique robot platforms instead of unique robots.

### DIFF
--- a/subt_ign/include/subt_ign/RobotPlatformTypes.hh
+++ b/subt_ign/include/subt_ign/RobotPlatformTypes.hh
@@ -25,17 +25,17 @@ const std::vector<std::string> robotPlatformTypes = {
   "X4",
   "HUSKY",
   "TEAMBASE",
-  "ANYMAL",
+  "ANYMAL_B",
   "M100",
   "FREYJA",
   "KLOUBAK",
   "HD2",
   "QAV500",
-  "EXPLORER_R2",
-  "EXPLORER_DS1",
-  "CSIRO_DATA61_OZBOT_ATR",
-  "CTU_CRAS_NORLAB_ABSOLEM",
-  "CERBERUS_GAGARIN"
+  "R2",
+  "DS1",
+  "OZBOT_ATR",
+  "ABSOLEM",
+  "GAGARIN"
 };
 
 #endif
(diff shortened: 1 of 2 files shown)


subt_ign/src/GameLogicPlugin.cc
```diff
@@ -767,7 +767,13 @@ void GameLogicPlugin::PostUpdate(
               auto filePath =
                 _ecm.Component<gazebo::components::SourceFilePath>(
                 model->Data());
-              this->dataPtr->robotSourceFilePaths.insert(filePath->Data());
+
+              // Store unique robot platform information.
+              for (const std::string &type : robotPlatformTypes)
+              {
+                if (filePath->Data().find(type) != std::string::npos)
```
iche033, line 774: some of the robot configurations are not all uppercase characters, e.g. `Robotika_Freyja_Sensor_Config_1`, so in this case, it will not match with the string `FREYJA`  in `robotPlatformTypes`. Maybe convert `filePath->Data()` to uppercase before doing the check?

subt_ign/include/subt_ign/RobotPlatformTypes.hh
```diff
@@ -0,0 +1,41 @@
...
+  "X4",
+  "HUSKY",
+  "TEAMBASE",
+  "ANYMAL",
+  "M100",
+  "FREYJA",
+  "KLOUBAK",
+  "HD2",
+  "QAV500",
+  "EXPLORER_R2",
```
angelacmaio, line 34: Should consistently use only the platform name, not the team name.
```suggestion
  "R2",
  "DS1",
  "OZBOT_ATR",
  "ABSOLEM",
  "GAGARIN"
```
* "

subt_ign/include/subt_ign/RobotPlatformTypes.hh
```diff
@@ -0,0 +1,41 @@
...
+/// \brief List of robot platform types. This is used to count unique robot
+/// types.
+const std::vector<std::string> robotPlatformTypes = {
+  "X1",
+  "X2",
+  "X3",
+  "X4",
+  "HUSKY",
+  "TEAMBASE",
+  "ANYMAL",
```
angelacmaio, line 28: Actual platform name is "ANYMAL_B"
```suggestion
  "ANYMAL_B",
```

subt_ign/src/GameLogicPlugin.cc
nkoenig: Fixed: e58ad453ee9ec1efed14e72304cbcf505db45d52

PR #555, opened 2020-08-18
Log breadcrumb events

nkoenig:
This adds breadcrumb deployments to the event log. This will capture breadcrumb deployment attempts.


nkoenig:
> The current functionality works for me.
> 
> It might make sense to add a `max_breadcrumb_deployments` event corresponding to a failed `breadcrumb_deploy` event that exceeds the limit of breadcrumbs deployed similar to how rock falls beyond the limit are recorded.

Add max deployment check in 8e74ef04e52ca914b55345d69770e9e0feed59eb

caguero (approved):
It works for me as expected as well.

PR #565, opened 2020-08-20
Path tracer improvements

nkoenig:
Support older log files, which may not have an `events.yml` .

Also fix a bug in the events.yml file which looks like:

```
- event:
  type: finished
  time_sec: 317
  elapsed_real_time 589
  elapsed_sim_time 205
  total_score: 1
```
Note the missing colons after `elapsed_real_time` and `elapsed_sim_time`.

### DIFF
--- a/subt_ign/src/path_tracer.cc
+++ b/subt_ign/src/path_tracer.cc
@@ -69,26 +69,61 @@ Processor::Processor(const std::string &_path, int _stepSleepMs)
   playbackThread.join();
 
   // Process the events log file.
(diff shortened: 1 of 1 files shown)


acschang (commented):
This update allowed for a successful run on a log file without an accompanying `events.yml` file.  I did not test for the formatting errors in the `events.yml` file.

iche033 (approved):
works for me. Tested with missing `events.yml` and one with bad syntax - the file now loads with the changes in this PR.

PR #566, opened 2020-08-21
Break out model download into separate docker image.

nkoenig:
This makes an new docker image that contains `ignition-blueprint` and the SubT models. This new image also acts as a base image for `cloudsim_sim` and `subt_sim_entry`.

The purpose for this is speed. If only code has changed, then we only need to rebuild `cloudsim_sim` and `subt_sim_entry` without re-downloading all of the models.

Usage of `cloudsim_sim` and `subt_sim_entry` does not change.


nkoenig:
Unnecessary change for right now.

PR #570, opened 2020-08-24
Spawner.rb syntax errors

acschang:
Resolves a syntax error preventing launch of the `ROBOTIKA_X2_SENSOR_CONFIG_1` model.  This syntax error is not present in any other submitted model.


jisa (commented):
Thanks for fixing it and even improving it on the way!

Issue #488, opened 2020-07-16
Communication strength with breadcrumbs


dsquez:
We have started experimenting with the breadcrumb functionality, and we are surprised by the apparent model for determining communication strength. New breadcrumbs appear to expand signal along line of sight (which makes sense), but this seems to be limited to within a certain radius of the staging area (or wherever the reference vehicle is when using the instructions provided in the wiki), which isn't what we would expect when setting up a communication system. I have created a google slides presentation that outlines where we put breadcrumbs and the subsequent changes in signal on Simple Cave 2. Does this follow the intended behavior, or might there be a bug?

https://docs.google.com/presentation/d/1tyE6HdsrkdCDjlSVKqXAX9lrhFQW6nzWBoz35OolC9Q/edit?usp=sharing

caguero:
Hi @dsquez , thanks for the great report. We're right now trying to reproduce your results. We'll go back to you soon.

caguero:
@dsquez , I can confirm that we found an issue with breadcrumbs. We're working on the fix.

malcolmst:
Just curious if there is any update on this issue? I haven't been experimenting much with the breadcrumbs yet but I see the PR for the fix is still open

nkoenig:
The PR is still open due to qualification code-freeze. We will be integrating communication updates next week.

malcolmst:
Makes sense, thanks for the update!

DalInar:
Hello! I've been experimenting with the breadcrumbs now that the updated comms model has been merged into master (#579, also related to #587 ), and it's behaving a bit differently than I (perhaps naively) expected.  Based on the new changes to the comms model, could you answer the following question:

What is the intended behavior of the RSSI value returned by the `getNeighbors` function when breadcrumbs are deployed?  In particular, say I have a TEAMBASE stationed in the starting area. Should I expect the RSSI between a robot and TEAMBASE to suddenly improve when a breadcrumb is deployed, because the RSSI is based on the connection between the breadcrumb and the robot, or should I simply expect the RSSI to fall off less quickly as I move away from the breadcrumb because the RSSI is somehow based on the entire path back to the starting area?

My simple experiments today (just manually moving a robot into the tunnel, manually deploying breadcrumbs, and then monitoring the RSSI returned by `getNeighbors`) seems to imply that the answer is the latter, but I was hoping for the former :-)

malcolmst:
[...]

PR #579, opened 2020-08-26
Updated comms model

caguero:
A pass of improvements over the communication model when breadcrumbs are deployed. 

The main change is a modification on how distance is computed in the model. Before this change, the Euclidean distance between source and destination was used even in the presence of breadcrumbs. This created a radial effect in the probability of message success as shown in the picture. This is inconsistent with the fact that breadcrumbs should extend the communication range. In the following example the source robot is in the staging area.

comms_v1

The new approach still takes into account distance but in a different way. Instead, we compute the greatest single hop of the route from source to destination. A hop is the distance from source to the first breadcrumb, between a pair of breadcrumbs, or between the last breadcrumb and the destination.  In the following example with the updated model the source robot is in the staging area.

comms_v2

The updated model also adds a small penalty (in distance) every time that a breadcrumb is added to the route.

This should close issue #488 .


acschang:
**For tracking sake:**
Example breadcrumb placement in PW01 is as follows:
pw01_v2_test1

There are three issues from this breadcrumb placement with this PR:
pw01_option_c_issues

1. Comms dead zone in green area
2. Comms dead zone between drop-off in the Y tile and the adjacent tile
3. Low comms escalating to high comms in further adjacent tile

These issues are not present with the current comms model (A) or visualization-only comms model (B):
pw01_legacy_bc_exp

subt_ign/src/VisibilityTable.cc
```diff
@@ -348,29 +378,109 @@ bool VisibilityTable::PopulateVisibilityInfoHelper(
...
+  const std::vector<ignition::math::graph::VertexId> &_relaySequence,
+  const ignition::math::graph::VertexId &_to) const
+{
+  if (_relaySequence.size() < 2)
+    return 0;
+
+  auto firstBreadcrumb = _relaySequence.at(0);
+  auto secondBreadcrumb = _relaySequence.at(1);
+
+  // We only consider the first breadcrumb stored in the tile.
```
acschang, line 467: Is it accurate that additional breadcrumbs in a tile have no bearing on the position max distance single hop?

acschang (commented):
Results look good on test cases with the resolution of all but the first issue (dead zone) which is not related to a source modification.  If you can clarify the comment I left on `VisibilityTable.cc` then I think we can be good to go on a merge here.

subt_ign/src/VisibilityTable.cc
caguero: > Is it accurate that additional breadcrumbs in a tile have no bearing on a max distance single hop from that tile?
> 
> e.g.
> Placing a breadcrumb at the beginning and end of a long tile would have an adverse effect on communication beyond the max hop distance would be unable to factor in the breadcrumb at the end of the long tile.

That's correct, we're not taking into account additional breadcrumbs within the same tile.

acschang: This could be related to https://github.com/osrf/subt/issues/587

angelacmaio:
Please update the [documentation](https://github.com/osrf/subt/wiki/Breadcrumbs-and-communication-visualization-tutorial) with more detail on how the comms model works once this is merged.

acschang (approved):
This is ready as issues 2 and 3 have been addressed.  A subsequent PR to address overlapping tiles blocking communication can be used to resolve issue 1 as that issue is not related to these changes and is present in the current communication model and the visibility-only communication model.

subt_ign/include/subt_ign/VisibilityTypes.hh
```diff
@@ -20,10 +20,39 @@
...
 {
+
+/// \brief A class that contains some information about the cost of
+/// communicating two robots.
+class VisibilityCost
+{
+  /// \brief The cost from one tile to another tile.
+  public: double cost;
+
+  /// \brief The best route connecting source and destination. No that route
```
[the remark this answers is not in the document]
caguero: Thanks, adf9ec9

PR #588, opened 2020-08-30
fix direction of lights on robot robotika/Freyja

m3d:
The lights unfortunately pointed to the ground - fixed now.
thanks
   Martin/Robotika Team

### DIFF
--- a/submitted_models/robotika_freyja_sensor_config_1/model.sdf
+++ b/submitted_models/robotika_freyja_sensor_config_1/model.sdf
@@ -1078,7 +1078,7 @@
         </material>
       </visual>
       <light name="light_front" type="spot">
-        <pose>0.237 0 0.0425 0 0 0</pose>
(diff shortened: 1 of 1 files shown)


nkoenig:
Lighting before:
freyja_before

Lighting after:
freyja_after


Looks good. This will make it into the code freeze.

m3d:
thanks a lot :)
m.

m3d:
Is there a way how to check that this was really deployed on CloudSim? We would like to fix also configuration 2 (after "unfreeze") and they both look the same??

nkoenig:
Those lighting issues are due to the javascript library used to render the scene. A fix has been applied [here](https://gitlab.com/ignitionrobotics/web/gz3d/-/merge_requests/33) and needs to be rolled out to the portal. The docker images do not require changes.

m3d:
@nkoenig great - thanks for info

PR #596, opened 2020-09-08
Path tracer rtf

nkoenig:
This changes the path tracer to use an RTF value when specifying playback speed instead of a time delay.

To run with RTF=1:

```
path_tracer <path_to_logs> 1.0
```

To run with RTF=2:
```
path_tracer <path_to_logs> 1.0
```

Need to update the wiki page: https://github.com/osrf/subt/wiki/Path-Tracer-Tutorial.

### DIFF
--- a/subt_ign/src/path_tracer.cc
+++ b/subt_ign/src/path_tracer.cc
@@ -238,6 +238,7 @@ void Processor::DisplayPoses()
   for (std::map<int, std::vector<std::unique_ptr<Data>>>::iterator iter =
        this->logData.begin(); iter != this->logData.end(); ++iter)
   {
+    auto start = std::chrono::steady_clock::now();
     printf("\r %ds/%ds (%06.2f%%)", iter->first, this->logData.rbegin()->first,
         static_cast<double>(iter->first) / this->logData.rbegin()->first * 100);
     fflush(stdout);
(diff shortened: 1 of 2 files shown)


subt_ign/src/path_tracer.cc
```diff
@@ -200,13 +200,25 @@ void Processor::ArtifactCb(const ignition::msgs::Pose_V &_msg)
...
     {
       data->Render(this);
     }
-    std::this_thread::sleep_for(std::chrono::milliseconds(this->stepSleepMs));
+
+    // Get the next time stamp, and sleep the correct amount of time.
+    auto next = std::next(iter, 1);
+    if (next != this->logData.end())
+    {
+      int sleepTime = ((next->first - iter->first) / this->rtf)*1000;
```
iche033, line 219: if we want to be accurate, we can set minus `sleepTime` by the amount of time it's already taken in this iteration. Not really necessary if the `for` loop with `Render` operation is relatively fast.

iche033 (approved):
Works for me.

subt_ign/src/path_tracer.cc
nkoenig: Good idea. I added it in here: b4adf4529d18f53a3793cc73984bc16dc22540a6.

PR #598, opened 2020-09-09
Fix report limit termination off by one error

nkoenig:
Missing `=` sign when comparing `reportCount` to `reportCountLimit`.

### DIFF
--- a/subt_ign/src/GameLogicPlugin.cc
+++ b/subt_ign/src/GameLogicPlugin.cc
@@ -1385,6 +1385,8 @@ bool GameLogicPluginPrivate::OnNewArtifact(const subt::msgs::Artifact &_req,
       << "  time_sec: " << this->simTime.sec() << "\n"
       << "  total_score: " << this->totalScore << std::endl;
     this->LogEvent(stream.str());
+    this->Finish();
+    return true;
   }
   else if (!this->ArtifactFromInt(_req.type(), artifactType))
   {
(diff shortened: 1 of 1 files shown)


subt_ign/src/GameLogicPlugin.cc
```diff
@@ -1432,11 +1434,20 @@ bool GameLogicPluginPrivate::OnNewArtifact(const subt::msgs::Artifact &_req,
...
+  if (!this->finished && this->reportCount >= this->reportCountLimit)
   {
-    _resp.set_report_status("report limit exceeded");
-    this->Log() << "report_limit_exceeded" << std::endl;
-    ignmsg << "Report limit exceed." << std::endl;
+    _resp.set_report_status("report limit reached");
+    this->Log() << "report_limit_reached" << std::endl;
+    ignmsg << "Report limit reached." << std::endl;
+
+    std::ostringstream stream;
```
mjcarroll, line 1443: This looks like it may duplicate logic from https://github.com/osrf/subt/pull/598/files#diff-8c34eb77e7a210f910eae0869c499eafL1380-L1387

Should the logging happen up there, and then here it will just `Finish` and `return True`?

mjcarroll (changes requested):

subt_ign/src/GameLogicPlugin.cc
nkoenig: This logic catches when report limit is reached (not exceeded). The logic on 1378 should never actually happen now, but it's safe to leave in.

mjcarroll (approved):

acschang (approved):
LGTM.  Reaches the limit then immediately sends summary and sets the status to finished.

PR #626, opened 2020-09-24
Event ids

nkoenig:
The goal is to have an ID that can be used to match ROS messages with events in the YAML file.

### DIFF
--- a/subt_ign/src/GameLogicPlugin.cc
+++ b/subt_ign/src/GameLogicPlugin.cc
@@ -207,6 +207,7 @@ class subt::GameLogicPluginPrivate
   /// \param[in] _simTime Current sim time.
   /// \param[in] _type Event type.
   /// \param[in] _robot Robot name.
+  /// \param[in] _eventId Unique ID of the event.
   public: void PublishRobotEvent(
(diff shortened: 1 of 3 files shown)


angelacmaio (approved):

subt_ign/src/GameLogicPlugin.cc
```diff
@@ -210,7 +210,8 @@ class subt::GameLogicPluginPrivate
   public: void PublishRobotEvent(
     const ignition::msgs::Time &_simTime,
     const std::string &_type,
-    const std::string &_robot);
+    const std::string &_robot,
+    int _eventId);
```
iche033, line 214: add `\param[in]` doxygen comment?

same for the function below

iche033 (commented):
looks good to me. Just one minor doxygen comment

iche033 (approved):

subt_ign/src/GameLogicPlugin.cc
nkoenig: 72498e9d89c6c06ebc146e58163b8a61c6203151

PR #640, opened 2020-09-30
Fix build error in subt_example (citadel branch)

adlarkin:
When I was using the `citadel` branch for communications visualization, the build failed on `subt_example` due to what appears to be an extra `add_subdirectory` call in [subt_example/CMakeLists.txt](https://github.com/osrf/subt/blob/c02b75337495e4e22851683a450a076cac9c086e/subt_example/CMakeLists.txt#L49). I have removed this call to fix the build - it looks like the `master` branch [doesn't use this `add_subdirectory` call](https://github.com/osrf/subt/blob/master/subt_example/CMakeLists.txt) either.

### DIFF
--- a/subt_example/CMakeLists.txt
+++ b/subt_example/CMakeLists.txt
@@ -46,7 +46,7 @@ add_executable(teleop_node src/teleop_node.cc)
 add_dependencies(teleop_node ${catkin_EXPORTED_TARGETS})
 target_link_libraries(teleop_node ${catkin_LIBRARIES})
 
-add_subdirectory(src/truth_controller)
+#add_subdirectory(src/truth_controller)
 
 
 ###########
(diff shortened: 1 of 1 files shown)


subt_example/CMakeLists.txt
```diff
@@ -46,7 +46,7 @@ add_executable(teleop_node src/teleop_node.cc)
 add_dependencies(teleop_node ${catkin_EXPORTED_TARGETS})
 target_link_libraries(teleop_node ${catkin_LIBRARIES})
 
-add_subdirectory(src/truth_controller)
+#add_subdirectory(src/truth_controller)
```
iche033, line 49: I also had to do this. I think you can just remove this line completely.

adlarkin: Removed in [d7abecb](https://github.com/osrf/subt/pull/640/commits/d7abecb65bc2e8598b1073d1b6041b3a753ea177).

iche033 (approved):

adlarkin:
It looks like @nkoenig already took care of this issue in [8b9d293](https://github.com/osrf/subt/commit/8b9d293ab208ba769f4d8258cf014049343c4768), so I am closing this PR.

PR #688, opened 2020-10-29
Do Release build instead of Debug.

zwn:
Should improve performance.

### DIFF
--- a/docker/cloudsim_bridge/Dockerfile
+++ b/docker/cloudsim_bridge/Dockerfile
@@ -110,7 +110,7 @@ WORKDIR /home/$USERNAME/subt_ws
 # RUN wget https://s3.amazonaws.com/osrf-distributions/subt_robot_examples/releases/subt_robot_examples_latest.tgz
 # RUN tar xvf subt_robot_examples_latest.tgz
 
-RUN /bin/bash -c 'source /opt/ros/melodic/setup.bash && catkin_make install'
+RUN /bin/bash -c 'source /opt/ros/melodic/setup.bash && catkin_make -DCMAKE_BUILD_TYPE=Release install'
 
 RUN /bin/sh -c 'echo ". /opt/ros/melodic/setup.bash" >> ~/.bashrc' \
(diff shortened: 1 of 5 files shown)


peci1:
Really? :-D

I've verified on a local build (without this PR, but also built just using `catkin_make install`):

```
$ cat build/subt/subt_ros/CMakeFiles/bridge_logger.dir/flags.make 
# CMAKE generated file: DO NOT EDIT!
# Generated by "Unix Makefiles" Generator, CMake Version 3.10

# compile CXX with /usr/bin/c++
CXX_FLAGS =   -std=c++11 -I/usr/include/uuid -std=gnu++1z

CXX_DEFINES = -DROSCONSOLE_BACKEND_LOG4CXX -DROS_BUILD_SHARED_LIBS=1 -DROS_PACKAGE_NAME=\"subt_ros\"

CXX_INCLUDES = -I/media/data/subt_ws/devel/include -I/media/data/subt_ws/src/subt/subt-communication/subt_communication_broker/include -I/media/data/subt_ws/src/subt/subt-communication/subt_communication_model/include -I/media/data/subt_ws/src/subt/subt-communication/subt_rf_interface/include -I/opt/ros/melodic/include -I/opt/ros/melodic/share/xmlrpcpp/cmake/../../../include/xmlrpcpp -isystem /usr/include/ignition/transport7 -isystem /usr/include/ignition/math6 -isystem /usr/include/ignition/cmake2 -isystem /usr/include/ignition/msgs4 -isystem /usr/include/uuid -isystem /usr/include/ignition/common3 
```

So the built libraries do not have debug symbols (`-g` is missing), but they also have no optimization (gcc defaults to `-O0` when no optimization is specified).

I'll test it in our builds.

zwn:
> Really? :-D

Yes. I was wondering why some of the bridge processes take around up to 8% of a 4GHz CPU not doing much of anything and I think this just might be it. I haven't done much testing yet but I thought I'd better publish it asap so others can benefit as well.

nkoenig:
I don't see any harm in this. I've update this PR in #762 .

PR #694, opened 2020-11-03
Update position of lights on side of Explorer X1 robot models

iche033:
The lateral lights on Explorer X1 models are found to be embedded inside the wheels. At certain camera viewing angles, the rendering engine thinks the lights are completely occluded and so the lights become disabled. This fix is to move the lights slightly outside the wheels so it does not get occluded. 

The red cubes in the screenshot below represent old pose of lights on the side of the robot:

explorer_light_pose

It was found that the issue is mostly in the GUI camera. As shown in the screenshot below, the light is not visible in the GUI camera but appears fine in the left camera sensor view:

explorer_x1_camera_sensor_view

Once this is merged, we'll also need to update the models on Fuel

### DIFF
--- a/submitted_models/explorer_x1_sensor_config_1/model.sdf
+++ b/submitted_models/explorer_x1_sensor_config_1/model.sdf
@@ -708,7 +708,7 @@
         <cast_shadows>1</cast_shadows>
       </light>
       <light name="lateral_left_front_headlight_body_light_source_light" type="spot">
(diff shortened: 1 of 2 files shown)


caguero (approved):

nkoenig (approved):

nkoenig:
I've also updated the models on Fuel.

PR #704, opened 2020-11-17
Read artifacts from world

nkoenig:
This removes the `<artifact>` tags from the launch files, and instead reads the set of available artifacts from the world itself.


mjcarroll:
Since we also do this in ArtfactValidator, I'm going to open a PR consolidating this logic.

mjcarroll:
> Since we also do this in ArtfactValidator, I'm going to open a PR consolidating this logic.

I'll do this in a follow-up.

mjcarroll (approved):

PR #705, opened 2020-11-17
Fix mean time calculation

nkoenig:
Off by one error in the calculation.


iche033 (approved):

subt_ign/src/GameLogicPlugin.cc
```diff
@@ -2580,7 +2580,7 @@ void GameLogicPluginPrivate::LogRobotArtifactData(
   if (!this->foundArtifacts.empty())
   {
     meanReportTime = (this->lastReportTime-this->firstReportTime) /
-        this->foundArtifacts.size();
+        (this->foundArtifacts.size() - 1);
```
azeey, line 2583: Can you add a check that this won't divide by 0?

nkoenig: Added in a2213c2112a958273299fc8660810793c10434c7

PR #708, opened 2020-11-19
Better bridge bagger

nkoenig:
This moves rosbag from launch into the SubtRosRelay node so that the bag file can be shutdown on `finish`. This switch makes sure that the bag file ends cleanly on Cloudsim.

### DIFF
--- a/subt_ros/launch/competition_init.launch
+++ b/subt_ros/launch/competition_init.launch
@@ -40,10 +40,11 @@
       name="set_pose_relay"/>
   </group>
 
+  <!-- A rosbag recorder is in the subt_ros/src/SubtRosRelay.cc -->
   <!-- rosbag a /robot_data topic which can be used by a team to record
-       custom data -->
+       custom data
   <node pkg="rosbag" type="record" name="rosbag_robot_data"
-        args="record --split --size=1000 --max-splits=2 -O $(env HOME)/.ros/robot_data.bag -e '/robot_data(.*)'"/>
+        args="record - -split - -size=1000 - -max-splits=2 -O $(env HOME)/.ros/robot_data.bag -e '/robot_data(.*)'"/>-->
 
   <node pkg="tf2_ros" type="static_transform_publisher" name="tf_world_static" args="0 0 0 0 0 0 world $(arg world_name)"/>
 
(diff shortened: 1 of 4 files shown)


subt_ros/launch/competition_init.launch
```diff
@@ -40,10 +40,11 @@
       name="set_pose_relay"/>
   </group>
 
+  <!-- A rosbag recorder is in the subt_ros/src/SubtRosRelay.cc -->
   <!-- rosbag a /robot_data topic which can be used by a team to record
-       custom data -->
+       custom data
   <node pkg="rosbag" type="record" name="rosbag_robot_data"
-        args="record --split --size=1000 --max-splits=2 -O $(env HOME)/.ros/robot_data.bag -e '/robot_data(.*)'"/>
+        args="record - -split - -size=1000 - -max-splits=2 -O $(env HOME)/.ros/robot_data.bag -e '/robot_data(.*)'"/>-->
```
mjcarroll, line 47: Nit: This could probably just be removed entirely at this point, since it will still be in the git history.

mjcarroll (approved):
LGTM.

subt_ros/src/SubtRosRelay.cc
```diff
@@ -222,6 +229,26 @@ SubtRosRelay::SubtRosRelay()
     this->rosnode->advertise<std_msgs::Int32>("score", 1000);
   this->rosCompetitionClockPub =
     this->rosnode->advertise<subt_ros::CompetitionClock>("run_clock", 1000);
+
+  // Setup a ros bag recorder.
+  rosbag::RecorderOptions recorderOptions;
+  recorderOptions.append_date=false;
+  recorderOptions.split=true;
+  recorderOptions.max_splits=1;
```
peci1, line 237: Before it was 2 :)

zwn: Yes, please don't make the logs smaller, thanks.

nkoenig: The previous implementation was incorrect. According to the [API](https://github.com/osrf/subt/wiki/api#custom-data-logging), up to 2 GB across 2 log files should be recorded. This equates to 1 split, which allows 2 log files.

This PR will produce up two 2 bag files, each with a maximum of 1GB.

subt_ros/launch/competition_init.launch
nkoenig: Removed in 753292ec9e79b397ed1197e9fd5ba45e60d31750

PR #726, opened 2020-12-16
Dome

nkoenig:
This will merge `dome` back into `master`, making Ignition Dome the new default for SubT.


nkoenig:
Dockerhub images have been updated to use this branch, and Cloudsim now also uses this branch. The wiki has also been updated to reflect the usage of Ignition Dome.

docker/subt_sim_entry/Dockerfile
```diff
@@ -82,7 +83,7 @@ RUN sudo apt-get update \
 # docker is run
 RUN mkdir -p subt_ws/src \
  && cd subt_ws/src \
- && git clone https://github.com/osrf/subt
+ && git clone https://github.com/osrf/subt -b dome
```
adlarkin, line 86: Should we clone the default branch instead of `dome` since I'm assuming the `dome` branch won't be updated anymore after merging this PR?

nkoenig: Fixed in 501f5a2120c98359e093df066d39e1e1cd8a6904

PR #732, opened 2020-12-17
Docker: Install ros-ign from Debian

nkoenig:
Install `ros-ign` from debian instead of from source so that the ros image compression library is brought in.

### DIFF
--- a/docker/cloudsim_bridge/Dockerfile
+++ b/docker/cloudsim_bridge/Dockerfile
@@ -34,7 +34,12 @@ RUN sudo apt-get update -qq \
         wget \
         net-tools \
         iputils-ping \
+        g++-8 \
  && sudo apt-get clean -qq
+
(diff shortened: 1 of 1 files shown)


mjcarroll (approved):

peci1:
Isn't this fix pretty fragile?

```
$ apt show ros-melodic-ros-ign 2>/dev/null | grep Depends
Depends: ros-melodic-ros-ign-bridge, ros-melodic-ros-ign-gazebo-demos, ros-melodic-ros-ign-image

$ apt show ros-melodic-ros-ign-gazebo-demos 2>/dev/null | grep Depends
Depends: libignition-gazebo4-dev, ros-melodic-image-transport-plugins, ros-melodic-ros-ign-bridge, ros-melodic-ros-ign-gazebo, ros-melodic-ros-ign-image, ros-melodic-rqt-image-view, ros-melodic-rqt-plot, ros-melodic-rviz
```

It only installs image transport plugins because `ros-melodic-ros-ign-gazebo-demos` depends on them. I guess that might never change, but why not install the plugins directly? Or why not explicitly use them as a dependency of ros-ign-image?

Issue #655, opened 2020-10-07
Breadcrumbs not being deployed


zwn:
In run `cb38d467-8aa8-483b-99ce-9e537a6e20fb` we have two robots deploying breadcrumbs - the names of the robots are `C600L` and `D40W600L`. According to our data each robot should have deployed 6 breadcrumbs, however only robot D40W600L deployed them (both robots run identical code). I was able to confirm that only D40W600L breadcrumbs are being deployed by inspecting `state.tlog` file. While trying to pinpoint the problem I was able to find all 6 lines containing `receiving: deploy []` in `rosout.log` for C600L which are being printed exactly before the ROS Empty message is published to the `/C600L/breadcrumb/deploy` topic. Further I can see in logfile `C600L-ros_ign_bridge_breadcrumbs-33-stdout.log` that this bridge has received the message at least once. Then in logfile `rostopic_stats_logger.log` I see 4 times the entry for `/C600L/breadcrumb/deploy` so at least 4 breadcumbs deploy commands have been received by `/C600L/ros_ign_bridge_breadcrumbs`.

To summarize: We published 6 times on the topic `/C600L/breadcrumb/deploy`, 4 times it has been received by `/C600L/ros_ign_bridge_breadcrumbs` but no breadcrumbs have been deployed.

zwn:
We see a similar thing happening in `46126b6f-1b8f-48a2-95a0-81e6e65db59b` - only the working and non-working robot have switched. I haven't done such a thorough analysis this time but I expect to find the same problem.

nkoenig:
This is just a status update. We are looking at the logs.

zwn:
Any progress on this? The deadline for cave is rapidly approaching (I bet you already knew that :wink:).

nkoenig:
We are still trying to reproduce this issue.

arjo129:
This happened to me on run `1e917331-cf3d-474b-a730-9d507ed787a0` one of my robots has definitely triggered breadcrumb deployments (`X1`) but it seems like the deployments did not work.

peci1:
In simulation e097b600-5ff5-4ce0-a687-8079f46d0815, this happened for our robot X3. Topic statistics show that the deploy commands were issued, but no breadcrumb from X3 appeared.

peci1:
But it doesn't happen often, this is actually the first time we noticed.

m3d:
X3? Is not X3 drone (without breadcrumbs)?

peci1:
[...]

PR #742, opened 2021-01-06
Enable topic statistics for breadcrumbs

nkoenig:
This turns on topic statistics for the breadcrumb plugins in order to help address issue #655 .

Requires:
1. https://github.com/ignitionrobotics/ign-gazebo/pull/532
2. https://github.com/ignitionrobotics/ign-launch/pull/81

When this is merged and deployed you will see the following in the Ignition Gazebo logs:

1. On successful transmission: `Received breadcrumb deployment for X1__breadcrumb__`
2. If a message was dropped: `Dropped message count of N for breadcrumbs on model X1__breadcrumb__`

To Test

1. `ign launch -v 4 competition.ign worldName:=simple_cave_01 circuit:=cave robotName1:=X1 robotConfig1:=X1_SENSOR_CONFIG_8`
2. `rostopic pub /X1/breadcrumb/deploy std_msgs/Empty`


subt_ros/launch/vehicle_topics.launch
```diff
@@ -1,5 +1,7 @@
 <?xml version="1.0"?>
 <launch>
+  <env name="IGN_TRANSPORT_TOPIC_STATISTICS" value="1"/>
```
azeey, line 3: Is this needed here if it's set in the `.ign` launch files? If so, we probably need to set it in each submitted model's `vehicle_topics.launch` file.

nkoenig: Yup, not needed and removed in 96e10d03b5a6a22cee47adcb847ba4faf3067749

subt_ign/launch/competition.ign
```diff
@@ -247,6 +247,11 @@
     <value>$LD_LIBRARY_PATH</value>
   </env>
 
+  <env>
```
azeey, line 250: nit: I'm not sure why this launch file has this. Shouldn't it be just in cloudsim_sim.ign?

subt_ros/launch/competition_init.launch
```diff
@@ -1,6 +1,8 @@
 <?xml version="1.0"?>
 <launch>
 
+  <env name="IGN_TRANSPORT_TOPIC_STATISTICS" value="1"/>
```
azeey, line 4: Probably not needed here as well.

azeey (approved):
Looks good. Just checking, do we want topic statistics on all models or just a subset? I only counted 10 spawner.rb files in this PR.

subt_ign/launch/competition.ign
nkoenig: I put this here so that the behavior is the same when running locally using this launch file and running in the cloud.

subt_ros/launch/competition_init.launch
nkoenig: I think this is needed in order get topic statistics on bridge, I'll double check.

PR #749, opened 2021-01-12
Call rosdep install from cloudsim dockerfiles and remove superfluous packages

peci1:
Extension of #748. Due to the use of rosdep install, some packages became superfluous.

I know this might become a breaking change to some users who assume some packages have already been installed. It shouldn't be that difficult for them to add the packages to their buildscripts (probably using rosdep again).

I rather split this PR apart from #748 so that if you choose to not go so aggresively with the optimizations, you can accept only #748 and reject this PR.

This PR makes sure things like RViz aren't installed in cloudsim images, as I don't see a real use for them there. It also does just a shallow clone of the subt repo and removes build artifacts and debug symbols from the built workspace.

These changes together save about 900 MB in the sim image and 800 MB in the bridge image (according to `docker system df -v`).

I did basic tests of the new images and everything seems to work as it should.


nkoenig:
@peci1 , can you update your branch with the latest changes on master?

peci1:
Updated.

peci1:
Another update with master (no conflict). I noticed the rosdep install commands were duplicated by the merge, so I only left the one command that was already there.

peci1:
I successfully built both of the images locally using this PR.

PR #751, opened 2021-01-13
More space-efficient subt_models Dockerfile

peci1:
This just moves all the space-wasting operations into a single Docker layer so that we do not needlessly store a layer with files which just get deleted a few steps further.

It also speeds up the installation by just installing ign-fuel-tools instead of whole ignition-dome.

Last, `apt-get autoremove` is called after uninstalling the temporary package, because otherwise, all of the packages installed as dependencies remain, which is probably unwanted.

This PR decreases total size of the models image from 5.5 GB to 4.5 GB. Taken into account that it is based on a 400 MB opengl image and downloads 3.6 GB of model data, the overhead is just 500 MB (most of which is an install of gcc-7 which is somehow pulled in but not uninstalled by autoremove... I didn't want to explicitly touch it).


nkoenig (approved):
Thanks. This looks good.

PR #755, opened 2021-01-19
Better support running from a local catkin workspace

peci1:
Currently, running a multirobot simulation (or multiple simulations) directly from local catkin workspace has its quirks. This PR addresses them.

1. robot_data bag files need to have a different name for different robots (and simulation partitions)... otherwise, all robots try to record to `~/.ros/robot_data_0.bag`, and only one of them obviously succeeds.
2. ignition log files (currently hardcoded to `/tmp/ign/logs`) need a configurable location, so that multiple simultaneous simulations do not write into the same folder.
3. The "simulator rosbag" needs to respect the path set in point 2. (currently, it hard-codes `/tmp/ign/logs` regardless of the `<path>` setting of GameLogicPlugin)

I think I saw some initiative towards 2. somewhere (passing a CLI arg with the path instead of having it hardcoded in the launch file), but I can't remember where it was.

Also, the parametrization of ign launch files with the log path would need to modify all of the provided files. I modified just cloudsim_sim.ign, and when we agree whether it's a good approach, I'd modify all of them.

The renaming of robot rosbags might break backwards compatibility for Docker users - the names of the files would change. I'm not sure if this is acceptable or not, but I can imagine some workarounds that would keep the original name if Docker environment is detected (e.g. matching `$USERNAME == developer`, or checking for nonempty `IGN_PARTITION`). Local catkin workspace users would also be affected, but as they were using broken logic anyways, it shouldn't actually hurt them.

### DIFF
--- a/subt_ign/launch/cloudsim_sim.ign
+++ b/subt_ign/launch/cloudsim_sim.ign
@@ -46,6 +46,12 @@
   else
     $circuit = circuit
   end
+
+  if !defined?(logPath) || logPath == nil || logPath.empty?
+    $logPath = '/tmp/ign/logs'
+  else
+    $logPath = logPath
+  end
 %>
 
 <%
@@ -340,7 +346,7 @@
     <record>
       <enabled>true</enabled>
       <!-- This path is used by cloudsim, please do not change -->
-      <path>/tmp/ign/logs</path>
+      <path><%= $logPath %></path>
       <overwrite>true</overwrite>
     </record>
     <%if defined?(seed) && seed != nil && !seed.empty?%>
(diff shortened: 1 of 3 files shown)


nkoenig (approved):

peci1:
This will need to be mentioned in the new release announcement. I suggest the following:

- `cloudsim_sim.ign` got a new parameter `logPath` which specifies where to record the simulator logs. Previously, it was hardcoded to `/tmp/ign/logs`.
- BAG files with robot data are named differently for each robot (and for each `IGN_PARTITION`). This allows running multirobot simulations from a local catkin workspace. For a robot with name `ROBOT`, the new name of the rosbags is as follows:
  - if `IGN_PARTITION` is unset or empty, the files will be named `robot_data_${ROBOT}_${i}.bag` (previously it was `robot_data_${i}.bag`).
  - If `IGN_PARTITION` is non-empty, the files will be named `robot_data_${IGN_PARTITION}_${ROBOT}_${i}.bag` (previously it was `robot_data_${i}.bag`).

Issue #665, opened 2020-10-14
Could we get transform world->artifact_origin?


peci1:
This transform would really be helpful in the solution containers. For example, we'd like to forbid exploration of the staging area, but its model is positioned relative to the world frame, which is inaccessible in the solution container. We can only guess that it will always be positioned at offset `10, 0, 0` from artifact_origin. Can we rely on this assumption? Or, even better, could we get the frame in the container?

peci1:
#756 addresses this issue.

PR #756, opened 2021-01-19
Publish static transforms describing the staging area in world frame.

peci1:
Solves #665.

Add publishing of the following static transforms to ROS:

- world->artifact_origin
- world->base_station
- world->staging_area

This is required e.g. to be able to precisely forbid exploration of the staging area after the robot left it. The relative positions of these three objects are not always the same, thus the transforms cannot be hard-coded.

Different from `/subt/pose_from_artifact`, this doesn't expose the Ignition services to the ROS interface, it just reads values from Ignition via service calls, and when the transform is known, publishes it as a static TF on the ROS side.

### DIFF
--- a/subt_ros/src/SubtRosRelay.cc
+++ b/subt_ros/src/SubtRosRelay.cc
@@ -32,6 +32,7 @@
 #include <subt_communication_broker/common_types.h>
 #include <subt_ros/CompetitionClock.h>
 #include <rosbag/recorder.h>
+#include <tf2_ros/static_transform_broadcaster.h>
 
 #include <ignition/transport/Node.hh>
 
@@ -149,6 +150,9 @@ class SubtRosRelay
   /// the origin artifact.
   public: ros::ServiceServer poseFromArtifactService;
 
+  /// \brief Static TF broadcaster to publish world->artifact_origin transform.
+  public: tf2_ros::StaticTransformBroadcaster staticTfBroadcaster;
+
   /// \brief The set of bound address. This is bookkeeping that helps
   /// to reduce erroneous error output in the ::Bind function.
   public: std::set<std::string> boundAddresses;
@@ -173,6 +177,9 @@ class SubtRosRelay
 
   /// \brief Pointer to the ROS bag recorder.
   public: std::unique_ptr<rosbag::Recorder> rosRecorder;
+
+  /// \bried Timer that tries to read the position of artifact origin in world.
+  public: ros::Timer artifactOriginPoseTimer;
 };
 
 //////////////////////////////////////////////////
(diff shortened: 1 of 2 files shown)


peci1:
This might benefit from using the `subt::kArtifactOrigin` etc. constants from subt_ign in the subt_ros part, but that would create a circular dependency.

There's also room for simplifying the code if I could add build dependency of subt_ros on ros_ign_bridge. That would allow using the `ign_to_ros()` conversion function. I'm not sure if that's desired, as the original pose_from_artifact code also doesn't utilize it.

mjcarroll:
I think we may need some clarification against the rules, since it looks like this would replace the use of `/subt/pose_from_artifact_origin`.  I believe the intention was to always make competitor robots use that service to get the `artifact_origin` location before leaving the staging area.

From the API page:

> The artifact pose should be reported relative to the artifact origin. At this point, there's no physical object to represent the artifact origin. Instead, robots can query the ROS service /subt/pose_from_artifact_origin that returns the robot pose relative to the artifact origin. Note that this service is only available if the robot is in the staging area.

peci1:
Well, if I understand it correctly, this shouldn't interfere with the rules. It doesn't add the possibility for a robot to query its own position in a different manner. What it adds is some "parent" or "neighbor" frames to `artifact_origin`. I'll try to explain why is which transform needed.

1. `staging_area`->`artifact_origin`: We want to be able to "mask out" the staging area from the areas allowed for exploration. If we don't know this, we have to somehow assume/derive the position of the area from the positions of the robots. That's mostly possible, but why not have it exactly. This would be needed both at competition and in testing.
2. `base_station`->`artifact_origin`: We'd want to know precisely where the base station is to have a better idea how to build the comms relay network, where to put the first breadcrumb and so on. Again, its position can be somehow estimated relative to the robot poses (or even detected visually), but it'd help to have the authoritative transform. Both at competition and in testing.
3. `world`->`artifact_origin`: This would only be helpful in testing when using ground truth. As the ground truth is published as transform `world`->`robot`, and all other transforms we have are `artifact_origin`->something, it's apparent that connecting these two subtrees is impossible without knowing the `world`->`artifact_origin` transform. In cave, it was the same for all worlds, but that doesn't hold generally. This transform would also be needed in case 1. and 2. are connected via the `world` frame (which they are as of now in this PR, but it might be possible to instead allow querying `staging_area` and `base_station` from `/subt/pose_from_artifact_origin`).

subt_ros/src/SubtRosRelay.cc
```diff
@@ -249,6 +256,45 @@ SubtRosRelay::SubtRosRelay()
...
+      t.transform.translation.z = pose.position().z();
+      t.transform.rotation.x = pose.orientation().x();
+      t.transform.rotation.y = pose.orientation().y();
+      t.transform.rotation.z = pose.orientation().z();
+      t.transform.rotation.w = pose.orientation().w();
+      this->staticTfBroadcaster.sendTransform(t);
+    }
+
+    if (successAll)
+      this->artifactOriginPoseTimer.stop();
```
adlarkin, line 295: So, if I understand this correctly, this timer will stop publishing the transforms once all of the transforms have been successfully defined/published in a single timer callback. My concern here is that this timer will stop publishing the transforms before the TF buffers on individual robots are initialized - if this happens, then a robot has no way to get these transforms, because these transforms wouldn't exist in the robot's TF buffer. Is a scenario like this possible?

cc @mjcarroll

peci1: This is okay, because all the transforms are published via `tf2_ros::static_transform_publisher` which internally creates a latched publisher with all transforms it has got throughout its existence. So no late joiners will miss no transforms.

peci1:
@adlarkin I've updated this PR with master (no conflifts).

This is the TF tree it produces:

frames

peci1:
I think merging this PR could still be desirable.

PR #758, opened 2021-01-19
BridgeLogger: Fix while(ros::ok) error.

peci1:
There's a coding bug in 

https://github.com/osrf/subt/blob/2af07e82fd2e67f7bb6bca267f6a26bcf1d75084/subt_ros/src/BridgeLogger.cc#L251

it checks that the address of function `ros::ok` is non-zero (which it is always). It should instead call the function.


adlarkin (approved):
Nice catch :grin:

peci1:
CLion catched it :-D

Issue #659, opened 2020-10-10
Ground Truth TF's slow after update


dan-riley:
Since updating to the latest Ignition and SubT versions (as detailed in the update page), enableGroundTruth is essentially broken.  We have tried several different robot models, and they all publish extremely slow TF's (around 0.5-1hz it seems) and there seems to be some other massive cpu or memory usage happening. When running with ground truth off everything runs as it did before.  I'm not sure which update caused the change, but we're seeing it on multiple machines.

nkoenig:
@dan-riley , did @acschang answer solve your problem?

dan-riley:
@nkoenig Yes, this seems to work.  We used 50hz.  I agree that it should probably be moved off of the static TF.

peci1:
Solution proposed in #759.

Issue #607, opened 2020-09-14
Ground truth pose publishers should not publish static TFs


peci1:
In all robot models, e.g.

https://github.com/osrf/subt/blob/9b8c41f85282c8b68f278dfc0a47e35148ba994b/submitted_models/ctu_cras_norlab_absolem_sensor_config_1/launch/spawner.rb#L107-L109

the ground truth pose is published as a static TF. It works, but it's not right, because static TFs should only be used for transforms that do not change or change only occasionally.

I think it could be fixed by adding this snippet to all robots:

```xml
    <plugin filename="libignition-gazebo-pose-publisher-system.so"
        name="ignition::gazebo::systems::PosePublisher">
        <publish_link_pose>false</publish_link_pose>
        <publish_sensor_pose>false</publish_sensor_pose>
        <publish_collision_pose>false</publish_collision_pose>
        <publish_visual_pose>false</publish_visual_pose>
        <publish_nested_model_pose>#{$enableGroundTruth}</publish_nested_model_pose>
        <use_pose_vector_msg>true</use_pose_vector_msg>
        <static_publisher>false</static_publisher>
      </plugin>
```

and setting 

```xml
    <publish_nested_model_pose>false</publish_nested_model_pose>
```

in the static transform snippets.

acschang:
Related to #659

peci1:
Solution proposed in #759.

iche033:
the changes look good to me and would resolve the issue that model poses are incorrectly published to the pose_static topic.

This makes me think that the PosePublisher system should really just be classifying a model / nested model entity as a dynamic entity instead of a static one (in which case, we would just insert the entity into the dynamic entities list somewhere around [here](https://github.com/ignitionrobotics/ign-gazebo/blob/ign-gazebo4/src/systems/pose_publisher/PosePublisher.cc#L402), i.e. `if (nestedModel) this->dynamicEntities.insert(entity)`)

PR #745, opened 2021-01-07
Combine all circuit launch files

adlarkin:
I have combined the following circuit launch files into a single file named `competition.ign`:
* [cave_circuit.ign](https://github.com/osrf/subt/blob/662cfffe7a1bee3be4c4b14700640a5179f419a1/subt_ign/launch/cave_circuit.ign)
* [tunnel_circuit_practice.ign](https://github.com/osrf/subt/blob/a9c45f1e5f15da76e261937fe7be17bb000427ed/subt_ign/launch/tunnel_circuit_practice.ign)
* [urban_circuit.ign](https://github.com/osrf/subt/blob/b73ab3fcdca5881504b8645ec73cdc5491c5d05b/subt_ign/launch/urban_circuit.ign)
* [competition.ign](https://github.com/osrf/subt/blob/a9c45f1e5f15da76e261937fe7be17bb000427ed/subt_ign/launch/competition.ign)

This new change has introduced a new parameter, `circuit`, that must be defined when running `ign launch`. `circuit` must be either `cave`, `tunnel`, or `urban`. Here are some example commands:

1. run a cave circuit world:
```
ign launch competition.ign circuit:=cave worldName:=simple_cave_01 robotName1:=X2_3 robotConfig1:=X2_SENSOR_CONFIG_3
```
2. run a tunnel circuit world:
```
ign launch competition.ign circuit:=tunnel worldName:=tunnel_circuit_practice_01 robotName1:=X1 robotConfig1:=X1_SENSOR_CONFIG_1
```
3. run an urban circuit world:
```
ign launch competition.ign circuit:=urban worldName:=urban_qual robotName1:=HUSKY robotConfig1:=COSTAR_HUSKY_SENSOR_CONFIG_1
```

---
**TODO** update wiki/READMEs anywhere that are affected by this change?

Signed-off-by: Ashton Larkin <ashton@openrobotics.org>

PR #759, opened 2021-01-19
Publish ground truth robot position as dynamic TF

peci1:
Solves #659 and #607. I hope I got all the places where ground truth publishing might be controlled.

Isn't there really a way to just `<include>` some common file defining all these common definitions for all robots? This approach seems pretty awkward to me...

### DIFF
--- a/subt_ign/launch/cloudsim_sim.ign
+++ b/subt_ign/launch/cloudsim_sim.ign
@@ -584,10 +584,20 @@
             <publish_sensor_pose>true</publish_sensor_pose>
             <publish_collision_pose>false</publish_collision_pose>
             <publish_visual_pose>false</publish_visual_pose>
-            <publish_nested_model_pose>#{$enableGroundTruth}</publish_nested_model_pose>
+            <publish_nested_model_pose>false</publish_nested_model_pose>
             <use_pose_vector_msg>true</use_pose_vector_msg>
             <static_publisher>true</static_publisher>
             <static_update_frequency>1</static_update_frequency>
+          </plugin>
+          <plugin filename="libignition-gazebo-pose-publisher-system.so"
+            name="ignition::gazebo::systems::PosePublisher">
+            <publish_link_pose>false</publish_link_pose>
+            <publish_sensor_pose>false</publish_sensor_pose>
+            <publish_collision_pose>false</publish_collision_pose>
+            <publish_visual_pose>false</publish_visual_pose>
+            <publish_nested_model_pose>#{$enableGroundTruth}</publish_nested_model_pose>
+            <use_pose_vector_msg>true</use_pose_vector_msg>
+            <static_publisher>false</static_publisher>
           </plugin>
           <!-- Battery plugin -->
           <plugin filename="libignition-gazebo-linearbatteryplugin-system.so"
@@ -726,10 +736,20 @@
             <publish_sensor_pose>true</publish_sensor_pose>
             <publish_collision_pose>false</publish_collision_pose>
             <publish_visual_pose>false</publish_visual_pose>
-            <publish_nested_model_pose>#{$enableGroundTruth}</publish_nested_model_pose>
+            <publish_nested_model_pose>false</publish_nested_model_pose>
             <use_pose_vector_msg>true</use_pose_vector_msg>
             <static_publisher>true</static_publisher>
             <static_update_frequency>1</static_update_frequency>
+          </plugin>
+          <plugin filename="libignition-gazebo-pose-publisher-system.so"
+            name="ignition::gazebo::systems::PosePublisher">
+            <publish_link_pose>false</publish_link_pose>
+            <publish_sensor_pose>false</publish_sensor_pose>
+            <publish_collision_pose>false</publish_collision_pose>
+            <publish_visual_pose>false</publish_visual_pose>
+            <publish_nested_model_pose>#{$enableGroundTruth}</publish_nested_model_pose>
+            <use_pose_vector_msg>true</use_pose_vector_msg>
+            <static_publisher>false</static_publisher>
           </plugin>
           <!-- Battery plugin -->
           <plugin filename="libignition-gazebo-linearbatteryplugin-system.so"
--- a/ign_migration_scripts/launch/preview.ign
+++ b/ign_migration_scripts/launch/preview.ign
@@ -209,10 +209,20 @@
             <publish_sensor_pose>true</publish_sensor_pose>
             <publish_collision_pose>false</publish_collision_pose>
             <publish_visual_pose>false</publish_visual_pose>
-            <publish_nested_model_pose>#{$enableGroundTruth}</publish_nested_model_pose>
+            <publish_nested_model_pose>false</publish_nested_model_pose>
             <use_pose_vector_msg>true</use_pose_vector_msg>
             <static_publisher>true</static_publisher>
             <static_update_frequency>1</static_update_frequency>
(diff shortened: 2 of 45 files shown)


subt_ign/launch/cloudsim_sim.ign
```diff
@@ -726,11 +736,21 @@
...
+          <plugin filename="libignition-gazebo-pose-publisher-system.so"
+            name="ignition::gazebo::systems::PosePublisher">
+            <publish_link_pose>false</publish_link_pose>
+            <publish_sensor_pose>false</publish_sensor_pose>
+            <publish_collision_pose>false</publish_collision_pose>
+            <publish_visual_pose>false</publish_visual_pose>
+            <publish_nested_model_pose>#{$enableGroundTruth}</publish_nested_model_pose>
+            <use_pose_vector_msg>true</use_pose_vector_msg>
+            <static_publisher>false</static_publisher>
+          </plugin>
```
adlarkin, line 753: There are 4 pose-publishers in this file (one for `X1`, `X2`, `X3`, and `X4`), but it looks like you've only updated 2 of them here. Would you mind updating the other 2?

adlarkin (commented):
Thanks for working on this! The changes here look good, but I'm not approving it yet because #745 combines a few of the circuit launch files into one - so, once #745 is merged, I think it would be good to re-base this on `master`, and then apply the ground truth update to just the newly combined `competition.ign` file instead of applying it separately to the launch files for cave, tunnel, and urban.

> Isn't there really a way to just <include> some common file defining all these common definitions for all robots? This approach seems pretty awkward to me...

This functionality should be available, but I'm not sure that we've taken advantage of it yet (I'd have to double-check). Perhaps we can make the `include` update in this PR, or in a separate PR after merging this.

peci1:
Okay, I can rebase on top of #745. Or do you expect it to change a lot until merged? And it'd be great to utilize some include facility, I can also add it to this PR. Is there some documentation/examples for it?

adlarkin:
> Okay, I can rebase on top of #745. Or do you expect it to change a lot until merged?

#745 hasn't been reviewed yet, so I'd wait to rebase in case any changes happen there. I don't think much will change, but perhaps reviewers for that PR will catch something that I missed originally. It's probably best to rebase this PR on `master` once #745 has been merged.

> And it'd be great to utilize some include facility, I can also add it to this PR. Is there some documentation/examples for it?

I don't think there is any documentation for this, but I have tested a possible solution with `X1_SENSOR_CONFIG_6`. Here's what you can do (if any of these steps seem confusing, keep in mind that the files we are dealing with are written in ruby):

1. Create a file that defines the pose publishing snippet. For example, you can put the following file in `subt/subt_ign/launch`: https://gist.github.com/adlarkin/eedb69675ead36373f2ddb6ab40fd6e8
2. In a submitted model's `spawner.launch`, make the following change (notice how the path is `../../subt_ign/...` instead of `../../../subt_ign/...` because this is the relative **install** path when using `catkin_make install`. `spawner.rb` is called from the install location when running `ign launch`, so we need to make sure that the paths are in the context of the workspace's `install/` directory). Here's an example with `x1_sensor_config_6`: https://gist.github.com/adlarkin/dc8ca333b154e8c98473918bbb8f9bed

I'm no ruby expert, so if you have a better idea for the `include` functionality, feel free to propose your solution here. If we do something similar to what I described above, then we will have to make sure that the ruby file with the pose publishing snippet is also included in the circuit launch files so that the "default" `X1`...`X4` configurations get this change as well (you'll also need to do something similar for [`ign_migration_scripts/launch/preview.ign`](https://github.com/osrf/subt/blob/1527a8ce093113cddf3a4615c13c6774e9a36fe5/ign_migration_scripts/launch/preview.ign#L205-L226)).

peci1:
Thanks for the hint, so I'll wait for #745. 

Isn't there a way to achieve the include via SDF? That would look more natural to me...

peci1:
And what would you think about extracting more pieces to the common include file - like gas sensor or joint state publisher?

adlarkin:
> Isn't there a way to achieve the include via SDF? That would look more natural to me...

I don't believe that there is. The `<include>` tag for SDF is meant for models: http://sdformat.org/tutorials?tut=composition&ver=1.5#including-a-model

So, we will probably need to do something with ruby... but maybe there's a cleaner solution than the one I proposed above.

> And what would you think about extracting more pieces to the common include file - like gas sensor or joint state publisher?

I think it would be worth using some sort of "include" mechanism for all common parts of model files. If we want to do something like this, it would probably be best to leave it for a separate PR, since this PR is really meant for fixing the topic for the ground truth data.

---

So, after some more thought, I'm thinking that the best approach moving forward would be this:
1. Make this PR fix the issue of publishing ground truth data to `/<name>/pose` instead of `/<name>/pose_static` (i.e., merge what you currently have implemented)
2. Open another PR that creates "include" functionality for the ground truth stuff done here, along with any other code snippets that are common across models

There are a few things to keep in mind:
1. #745 combines some of the launch files - after some more thought, I think it's fine to go ahead and update all of the individual circuit launch files with the ground truth fix in this PR. I can always update #745 later if this PR gets merged first (and if #745 gets merged before this, then we can update this PR instead... we'll just figure it out when the time comes)
2. There are several other newly submitted models that are currently in review/being merged... we probably want to wait to merge this PR until all of the other submitted models are merged. Once the other submitted models are merged, we should then update this PR with the newly merged models, and add this fix to the new models.

What do you think about this approach? Any thoughts/objections?

peci1:
I updated this PR with the changes from #745 and also applied the changes to the newly accepted robots, so it's ready to be merged.

subt_ign/launch/cloudsim_sim.ign
peci1: Good catch, updated.

peci1:
I opened up #786 with the larger refactoring. Could you please have a look, @adlarkin ?

adlarkin:
> I updated this PR with the changes from #745 and also applied the changes to the newly accepted robots, so it's ready to be merged.

There were a few more robot configurations added in the last week or two that might need these updates as well:
* https://github.com/osrf/subt/pull/779
* https://github.com/osrf/subt/pull/777
* https://github.com/osrf/subt/pull/773

> I opened up #786 with the larger refactoring. Could you please have a look, @adlarkin ?

Yeah, I'll take a look (sorry for the delayed response). If #786 is based off of this PR, do we still need this PR? Or can we close it?

nkoenig:
> Yeah, I'll take a look (sorry for the delayed response). If #786 is based off of this PR, do we still need this PR? Or can we close it?

I think it's safe to get this PR in, and then @peci1  can update #786.

peci1:
> I think it's safe to get this PR in, and then @peci1 can update #786.

Yes, that's the preferred option. I'll update this PR later today.

peci1:
Updated. I did a search throughout whole repo and all files except validate_connections.ign show an even number of occurences of `ignition-gazebo-pose-publisher-system`.

adlarkin (commented):
Changes look okay to me. @nkoenig can you or someone else take a quick look at this to make sure we didn't miss any updates anywhere?

peci1:
I've updated this PR with the latest changes from master (no merge conflicts).

peci1:
I think this PR would still bring value in the simulator. However, it'd need a major update since it was last synced with master in March... Is the idea of this PR still welcome?

peci1:
I've updated this PR so that it handles all models currently in `master`.

PR #760, opened 2021-01-19
Publish ground truth data to /<name>/pose instead of /<name>/pose_static

### LINKED ISSUE #607: Ground truth pose publishers should not publish static TFs

### LINKED ISSUE #659: Ground Truth TF's slow after update

adlarkin:
Currently, ground truth pose data is being published to a robot's `pose_static` topic, which is problematic for the following reasons:
1. This implies that the ground truth pose doesn't change (or doesn't change often, at least), which is not true for most use cases.
2. The `pose_static` topic publishes at a rate of 1Hz, making the ground truth data difficult to use for real-time debugging (the `pose` topic publishes at a rate of 250Hz).

Resolves #659 and resolves #607.

TODO:
- [ ] Make models X1 ... X4 also publish ground truth data to `/<name>/pose` (I'm waiting for #745 to be merged before doing this)

### DIFF
--- a/submitted_models/cerberus_anymal_b_sensor_config_1/launch/spawner.rb
+++ b/submitted_models/cerberus_anymal_b_sensor_config_1/launch/spawner.rb
@@ -45,10 +45,21 @@ def spawner(_name, _modelURI, _worldName, _x, _y, _z, _roll, _pitch, _yaw)
         <publish_sensor_pose>true</publish_sensor_pose>
         <publish_collision_pose>false</publish_collision_pose>
         <publish_visual_pose>false</publish_visual_pose>
-        <publish_nested_model_pose>#{$enableGroundTruth}</publish_nested_model_pose>
+        <publish_nested_model_pose>false</publish_nested_model_pose>
         <use_pose_vector_msg>true</use_pose_vector_msg>
         <static_publisher>true</static_publisher>
         <static_update_frequency>1</static_update_frequency>
+      </plugin>
+      <!-- Publish ground truth information if specified by user -->
+      <plugin filename="libignition-gazebo-pose-publisher-system.so"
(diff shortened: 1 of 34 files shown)


adlarkin:
Closing this because the same solution was proposed in #759

PR #762, opened 2021-01-20
Update to pr #688

nkoenig:


peci1:
Would you also update https://github.com/osrf/subt/wiki/Catkin%20System%20Setup ?

PR #767, opened 2021-01-25
Added more logging to the shutdown process

nkoenig:
The goal is to make it a little easier to debug the shutdown process.

### DIFF
--- a/subt_ign/src/GameLogicPlugin.cc
+++ b/subt_ign/src/GameLogicPlugin.cc
@@ -2272,7 +2272,10 @@ void GameLogicPluginPrivate::Finish(const ignition::msgs::Time &_simTime)
         << "  elapsed_sim_time: " << simElapsed << "\n"
         << "  total_score: " << this->totalScore << std::endl;
       this->LogEvent(stream.str());
+      this->eventCounter++;
+    }
 
+    {
       // \todo(nkoenig) After the tunnel circuit, change the /subt/start topic
       // to /sub/status.
       ignition::msgs::StringMsg msg;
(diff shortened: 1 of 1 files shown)


subt_ign/src/GameLogicPlugin.cc
```diff
@@ -2307,6 +2317,17 @@ void GameLogicPluginPrivate::Finish(const ignition::msgs::Time &_simTime)
...
+        << "- event:\n"
+        << "  id: " << this->eventCounter << "\n"
+        << "  type: recording_complete\n"
+        << "  time_sec: " << _simTime.sec() << "\n"
+        << "  elapsed_real_time: " << realElapsed << "\n"
+        << "  elapsed_sim_time: " << simElapsed << "\n"
+        << "  total_score: " << this->totalScore << std::endl;
+      this->LogEvent(stream.str());
       this->eventCounter++;
     }
```
AlejoAsd, line 2332: Could we add an `else` case to mention that the ROS node is not active? That way we can verify that the block has not been accessed.

nkoenig: the `if (this->rosnode)` block ends on line 2310. The new event starting on line 2320 will always output, regardless of ROS usage. That event will be logged in `events.yml`.

AlejoAsd: Ah, my bad. My eyes are used to larger indents, or maybe I need some glasses. Thanks!

AlejoAsd (approved):

subt_ign/src/GameLogicPlugin.cc
AlejoAsd: Ah, my bad. My eyes are used to larger indents, or maybe I need glasses. Thanks!

mjcarroll (approved):

Issue #638, opened 2020-09-30
CommsClient: Binding from multiple nodes doesn't work (even on different ports).


peci1:
I've found no mention in documentation that there can be only a single node using CommsClient::Bind() with a given address, no matter the port number. But that is the reality. Further in this issue is my argumentation why this should be considered a bug (either in CommsClient implementation or in documentation).

This is the code executed in CommsClient's OnBind() method:

https://github.com/osrf/subt/blob/2f63e31e4960e49680d09dd66f7cbea8a122fa07/subt-communication/subt_communication_broker/src/subt_communication_client.cpp#L243-L244

It clearly shows that it creates a service server whose name doesn't depend on port number. If your robot's name is `X1`, then the service server is `/X1` no matter what port you bind to. The code checks that the server is only started once per CommsClient instance, and messages from all ports are gathered by this CommsClient. It then redistributes the messages to the bound clients. But if I start another node with another instance of CommsClient, the service is already advertised and thus the new instance cannot listen for new messages, even if it would listen on different ports than the first instance.

I've spent hours digging into this bug and have come to several thoughts:

- `/$ADDRESS` is the worst name one could imagine... it makes sense to me to run robots in namespaces according to their name, and this effectively makes it impossible to remap just the CommsClient's service, because all topics on our robots start with the same prefix. What about `/subt/comms/$ADDRESS`, `/comms/$ADDRESS` or similar?
- Why is it a service at all? 

https://github.com/osrf/subt/blob/d5f8fe760cd6f0e68a6a9a615f143b74d82d07dc/subt_ros/src/SubtRosRelay.cc#L466-L476

The SubtRosRelay code calls the services, but doesn't care about their result (or even success). Moreover, if some client's processing of the message takes longer, it will block the whole message relay. If it were a standard ROS topic, it would not block and it would even allow multiple binds to the same address:port pair (which is not normally supported with UNIX ports, but could be nice).
- From the previous point, if multiple binds of the same address:port pair are not allowed, then the user should be allowed to choose which CommsClient will receive the beacon packets. Because so far, the bind for beacons is done automatically in the CommsClient's constructor, so the first one that gets created wins. Not really practical. And I think this would be a very good reason to support multiple binding.
- (slightly unrelated) The documentation (and comms tutorial) do not explicitly mention the port on which the base station communicates (the default port). It could help getting a clearer understanding of the reporting API.

[...]

Issue #604, opened 2020-09-11
Multicast over SubT comms is not allowed


tpet:
I am not able to bind to a multicast group.
E.g. the code below (excerpt from within a member function)
```
comms_client_ = std::make_shared<subt::CommsClient>("X1");
// subt::communication_broker::kMulticast == "multicast"
comms_client_->Bind(&SimpleRelay::comms_callback, this, subt::communication_broker::kMulticast, 20) // returns false
```
fails with
`[ERROR 1599819990.746046087 /subt_ros_relay] OnBind address does not match origination. Attempted impersonation of a robot as robot[multicast].`

I'm using an image based on osrf/subt-virtual-testbed:latest.

Is multicast supported for cave circuit? From [the comms client docs](https://github.com/osrf/subt/blob/master/subt-communication/subt_communication_broker/include/subt_communication_broker/subt_communication_client.h#L100) it seems that it should be. (`"kMulticast"` doesn't work either.)

AravindaDP:
@nkoenig @caguero Any updates on this? I also came across same issue and couldn't figure out how to properly use multicast functionality. Could there be a bug in underline CommsClient?

peci1:
Solution proposed in #772.

Issue #270, opened 2019-11-12
CommsBroker unregister being called?


osrf-migration:
**Original report ([archived issue](https://osrf-migration.github.io/subt-gh-pages/#!/osrf/subt/issues/270)) by Steven Gray (Bitbucket: [stgray](https://bitbucket.org/%7B7470e809-c304-4b17-b986-685f31fa15c3%7D/)).**

----------------------------------------
This looks very much like ~~#33~~. I’m working on my comms node and would like to restart it multiple times without restarting the sim.

To recreate - repeated running of the `subt_example_node` without also restarting the sim also results in this failure:

```
Broker::Register() warning: ID [X1] already exists
Broker::Bind() error: Address [X1] already used in a previous Bind()
```

Produced with the following – including killing and restarting `subt_example_node` once.

```
ign launch -v 4 urban_circuit.ign worldName:=urban_qual robotName1:=X1 robotConfig1:=X2_SENSOR_CONFIG_6
rosrun subt_example subt_example_node X1
```

‌

tpet:
We have same problems - after restarting the comms client, it cannot bind to the same address and transmitting messages over subt comms doesn't work anymore.

Maybe a hint - couldn't subscribing to multiple ports cause the problems (e.g., that only one port is unregistered when the client is destroyed)?

I would also propose to re-label this as a major bug instead of minor proposal.

nkoenig:
Thank you for the suggestion. A bug is an issue that prevents the SubT Simulator from operating correctly or blocks development. Restarting robot code without restarting simulation is not a supported use case at this time. Therefore this issue is an enhancement.

peci1:
Solution proposed in #772.

PR #772, opened 2021-02-01
Add support for running multiple CommsClient nodes

peci1:
This PR adds proper support for having multiple CommsClients in different ROS nodes (or even multiple clients in one ROS node).

Solves: #638, #604 (yes, multicast!) and #270.

It is a rather large PR. I must admit much larger than I imagined when I started to work on it (given the fact that our half-sufficient workaround was just one file; but that workaround still did not work when nodes got restarted). Yet I still think this is the (more or less) minimal set of changes to provide proper support for multiple clients at least on the ROS side. There can still be only a single CommsClient with `_useIgnition=true`. But that's not a big problem because that client is a part of the bridge (or sim) container, so it's easy to make sure there is only one. On the other hand, solution containers using `_useIgnition=false` will now get much bigger freedom in spawning commsclients.

I even added support for calling `Unbind()`, although it seems to me that it's not that useful in SubT challenge (but hey, what if!).

## What this PR brings

1. Creating multiple instances of `subt::communication_client::CommsClient` (even in one ROS node, e.g. several nodelets can have their own commsclients). Each of these clients has its own separate list of bound endpoints which get unregistered when this client goes out of scope.
2. Binding one endpoint multiple times even from the same client instance (e.g. to add multiple independent processing callbacks).
3. Unbinding endpoints by calling `CommsClient::Unbind()`. This will stop calling the callback bound in the corresponding `Bind()` call.
4. Restarting ROS nodes with CommsClients. This was not possible before because each endpoint could only be bound once and there was no unbind option. **This is a big step towards reliability and robustness** of the solution containers - i.e. the important commsclient nodes can finally get `respawn="true"` and they will still work even after a crash.
5. The execution of the bound callbacks of one client does not block/delay the execution of callbacks of other clients (because they are dispatched via a ROS topic instead of service).
6. Multicast addresses work as they should.

## What this PR breaks

This PR of course breaks a lot of APIs/ABIs. I tried to keep the changes for downstream users to a minimum, and a proof that I was successful in that is that there's literally no "downstream" code in osrf/subt that would need any changes (BaseStation/Controller plugins, tests, subt_example_node). Here's a view of the changes from a few different perspectives.

### Users of `subt::communication_client::CommsClient(..., _useIgnition=false)`

1. The only change is the return value of `Bind()` call which now returns a list of registered endpoint IDs instead of plain bool. I added extra backwards compatibility feature by overloading `operator!` for the result type, so that people can still easily test `if (!client->Bind(...)) ...`. I couldn't figure out how to support also a positive test like `if(client->Bind(...))...`. People who use just the CommsClient and did not check the return value of Bind() should be completely unaffected by this PR (except gaining all of the bonuses it brings).
2. The constructor of `CommsClient` has two more optional parameters.
 - `_listenBeacons` specifies whether this client should automatically bind to the beacon port and fill the Neighbors() array
 - `_rosNh` is an optional ROS nodehandle to use for the created/subscribed ROS topics/services (e.g. in case users want to remap some of the topics)

### Direct users of the ROS API:

1. Definitions of the service calls in `subt_ros/srv` have changed. They're now using client/endpoint IDs instead of plain names for unregistration, unbind and as return values. The Bind.srv also requires a related client ID instead of address. It should be pretty easy to maintain a mapping between client/endpoint IDs and their corresponding names.
2. Incoming messages from the SubRosRelay no longer come via a service call to `/robot`, but are instead published on topic `/robot/comms`. This is the key part that allows multiple CommsClients to coexist.

### Users of `subt::communication_client::CommsClient(..., _useIgnition=true)`

1. As with `_useIgnition=false`, there is nothing changed but the return type of `Bind()`.
2. This client instance does not, however, support multiple instances running at the same time (although creation of additional clients doesn't fail; they will just not receive any messages).

### Direct users of the Ignition Transport API:

1. Definitions of the registration/binding services changed. See `Broker::OnAddrRegistration`, `Broker::OnAddrUnregistration` and `Broker::OnEndpointRegistration` for their new signatures. All `UInt32` fields stand for some ID (clientId or endpointId).

## Validation

To be more sure this PR actually does what it says, I added lots of unit/integration tests. They can probably still be extended, but I think they capture the basic functionality well. As another measure to verify validity of the changes, I ran a subt_comms_test with two subt_example robots, and everything went fine. I was even able to restart each of the "solution" launch files without restarting the simulator and the test still went ok (but it needed subt_comms_test to be restarted because it lost track of the connections).

As a last point, I want to run a full systems test with our solution containers and a full 6-robot team. I will exchange our custom multi-client workaround for the changes in this PR and observe the behavior of the system. Our system normally runs two commsclients on the teambase - on for communicating with other robots, and one for communication with the basestation. So this test should practically verify that a multi-client setup works. I will also try to kill some of the communication nodes to see whether restarting them is possible and they are fully operational afterwards.


peci1:
I successfully ran a full 3-robot test with both simulator and our solution code switched to use this PR

- [x] The robots successfully communicated with each other and with teambase.
- [x] UGVs were each running two commslients (each in a separate ROS node), and TEAMBASE was running 3 commsclients (each in separate ROS node).
- [x] Both broadcast and unicast was used for the communication.
- [x] Hundreds of endpoints were bound during the test without an issue.
  - Many fast concurrent binds discovered a small concurrency issue which I fixed.
- [x] Communication with base station and artifact reports went through smoothly.
- [x] I even tried restarting the nodes with commsclients "on-the-fly" and they happily reconnected and started sending/receiving messages after the restart.

peci1:
Updated with master (no conflict).

PR #776, opened 2021-02-02
Fix thermal camera topic mapping

iche033:
changes so it works with the current thermal camera topic names

* topic: no `/image` suffix
* camera_topic: incorrectly crops sensor name in topic string

fixing the topic names in ign-gazebo breaks existing behavior so using this workaround for now.


submitted_models/marble_hd2_sensor_config_1/launch/vehicle_topics.launch
```diff
@@ -257,19 +257,21 @@
...
+    <!-- Note camera_info topic generated by ign-gazebo incorrectly removes
+         sensor name in topic string -->
     <node
       pkg="ros_ign_bridge"
       type="parameter_bridge"
       name="ros_ign_bridge_thermal_camera"
-      args="$(arg sensor_prefix)/thermal_camera/camera_info@sensor_msgs/CameraInfo[ignition.msgs.CameraInfo">
-      <remap from="$(arg sensor_prefix)/thermal_camera/camera_info" to="thermal_camera/camera_info"/>
+      args="$(arg sensor_prefix)/camera_info@sensor_msgs/CameraInfo[ignition.msgs.CameraInfo">
+      <remap from="$(arg sensor_prefix)/camera_info" to="thermal_camera/camera_info"/>
```
adlarkin, line 267: Is it safe to say that the thermal camera will be the only camera who's `camera_info` topic is `$(arg sensor_prefix)/camera_info`? I know that we need this workaround since we can't change the topic in `ign-sensors`, but I just want to make sure that there will never be any other kind of cameras that would collide with this topic.

adlarkin (approved):
Looks good, thanks for the fixes! Just one minor question below.

Also, do we need the optical frame publishers for the thermal camera, or can we leave them out?

submitted_models/marble_hd2_sensor_config_1/launch/vehicle_topics.launch
iche033: yeah luckily the, normal camera, rgbd camera, and depth camera do not have this problem so they do not conflict with this `camera_topic` topic string here

iche033:
> Also, do we need the optical frame publishers for the thermal camera, or can we leave them out?

Since they are also cameras that produce `sensor_msgs/Image` data so we follow [this ROS REP](https://www.ros.org/reps/rep-0103.html) and provide an optical frame. I'm not sure if teams will use it though.

PR #797, opened 2021-02-19
Adjusted temperature variation per world

angelacmaio:
Adjusted ambient (in-scene) temperature variation per world to better represent contrast across different environments.

Side note: the `<temperature_gradient>` parameter does *not* cause ambient temperature to change by altitude as the [SDFormat specs ](http://sdformat.org/spec?ver=1.7&elem=world#atmosphere_temperature_gradient) suggest.


iche033:
Looking through the temperature gradients in the world, the values are from 0.2 to 0.7.

Here's a mapping of temperature gradient to the computed temperature range:

gradient (K/m) -> range (K)
0.2   -> 10.0477
0.3   -> 15.0715
0.4   -> 20.0953
0.5   -> 25.1192
0.6   -> 30.143
0.7   -> 35.1668

if these values are fine then changes look good to me.

angelacmaio:
@iche033 those values are fine. It seems like a very large variation, but the thermal camera output looks reasonable especially when comparing environment variation to artifact temps. Thanks!

iche033 (approved):

Issue #808, opened 2021-03-04
Finals Qualification World - failing to report artifact


m3d:
Hello,
was there any change we should be aware since "Cave Circuit" related to artifact reporting? We are not able to report any (even false reports) and in the internal logs we see that some were reported??
thanks
   Martin/Robotika Team

m3d:
in the rosout I see
```
243.552000000 INFO /A300L_proxy [/osgar-ws/src/osgar/subt/ros/proxy/ros_proxy_node.cc:569(receiveZmqThread)] [topics: /rosout, /A300L/cmd_vel, /statistics] artf: artf TYPE_BACKPACK 35.83 28.94 1.22

243.552000000 INFO /A300L_proxy [/osgar-ws/src/osgar/subt/ros/proxy/ros_proxy_node.cc:447(msgs::Artifact&))] [topics: /rosout, /A300L/cmd_vel, /statistics] MD artf = TYPE_BACKPACK
243.552000000 INFO /A300L_proxy [/osgar-ws/src/osgar/subt/ros/proxy/ros_proxy_node.cc:448(msgs::Artifact&))] [topics: /rosout, /A300L/cmd_vel, /statistics] MD x = 35.83
243.552000000 INFO /A300L_proxy [/osgar-ws/src/osgar/subt/ros/proxy/ros_proxy_node.cc:449(msgs::Artifact&))] [topics: /rosout, /A300L/cmd_vel, /statistics] MD y = 28.94
243.552000000 INFO /A300L_proxy [/osgar-ws/src/osgar/subt/ros/proxy/ros_proxy_node.cc:450(msgs::Artifact&))] [topics: /rosout, /A300L/cmd_vel, /statistics] MD z = 1.22
243.552000000 INFO /A300L_proxy [/osgar-ws/src/osgar/subt/ros/proxy/ros_proxy_node.cc:516(msgs::Artifact&))] [topics: /rosout, /A300L/cmd_vel, /statistics] MD enum = 0
243.552000000 DEBUG /A300L_proxy [/tmp/binarydeb/ros-melodic-roscpp-1.14.9/src/libros/transport/transport_tcp.cpp:142(TransportTCP::initializeSocket)] [topics: /rosout, /A300L/cmd_vel, /statistics] Adding tcp socket [49] to pollset
[...]

PR #809, opened 2021-03-05
Update comms model to cover base station position

iche033:
Here's a visualization of the new comms model. If you look at the staging area, the green points clouds are now extended (in -x) to cover the position of the base station model which was added in https://github.com/osrf/subt/pull/806

comms_base_station

Note the green points maybe hard to see over white background. You may need to zoom out a bit to see the visualizations


angelacmaio (approved):
LGTM. Resolves #808

acschang (commented):
LGTM.

PR #816, opened 2021-03-11
Fix segfault when a robot with flippers is a part of the world at startup

peci1:
When I create a simple world with CTU_CRAS_NORLAB_ABSOLEM_SENSOR_CONFIG_1 included using the `<include>` SDF tag, gazebo segfaults. I tracked it down to the pose component being present but empty (probably a startup sequence issue, that's why it didn't manifest in SubT, because the robots are spawned into an existing world, as opposed to my case when the robot is a part of the world from the beginning). This PR adds additional checks to make sure empty pose component is skipped.


iche033 (approved):
looks good to me

Issue #412, opened 2020-05-04
Cloudsim startup and total runtime vs simulation time


zwn:
I experiencing a couple of strange behaviors on cloudsim.

First, I am still fighting with how to know when everything is ready to start. First suggestion was to wait for /clock but that is not enough (even when /clock is available, other stuff is not). At the end, I am waiting almost for everything the solution might need, one by one. This is an example or rosout from one such a run (it is short so I include it in full):

```
0.000000000  Node Startup
95.872000000 INFO /rosbag_robot_data [/tmp/binarydeb/ros-melodic-rosbag-1.14.5/src/recorder.cpp:394(Recorder::startWriting)] [topics: /rosout] Recording to '/home/developer/.ros/robot_data_0.bag'.
0.000000000 INFO /tf_world_static [/tmp/binarydeb/ros-melodic-tf2-ros-0.6.5/src/static_transform_broadcaster_program.cpp:142(main)] [topics: /rosout, /tf_static] Spinning until killed publishing world to simple_urban_01
0.000000000 DEBUG /cudatest [tcpros_base.py:559(TCPROSTransport.connect)] [topics: /clock, /rosout] connecting to 10.44.0.2 52869
341.904000000 INFO /cudatest [entrypoint.py:17(main)] [topics: /clock, /rosout] got rosout logger
341.904000000 INFO /cudatest [entrypoint.py:19(main)] [topics: /clock, /rosout] waiting for clock
341.908000000 INFO /cudatest [entrypoint.py:21(main)] [topics: /clock, /rosout] rospy.Time[341908000000]
341.908000000 DEBUG /cudatest [tcpros_service.py:112(contact_service)] [topics: /clock, /rosout] connecting to ('10.44.0.2', 38949)
341.912000000 DEBUG /cudatest [tcpros_service.py:112(contact_service)] [topics: /clock, /rosout] connecting to ('10.44.0.2', 38949)
341.912000000 DEBUG /cudatest [tcpros_base.py:559(TCPROSTransport.connect)] [topics: /clock, /rosout] connecting to 10.44.0.2 38949
341.920000000 INFO /cudatest [entrypoint.py:27(main)] [topics: /clock, /rosout] /subt/start: True
344.384000000 INFO /cudatest [entrypoint.py:32(main)] [topics: /clock, /rosout] torch ok
344.388000000 INFO /cudatest [entrypoint.py:32(main)] [topics: /clock, /rosout] tf ok
[...]

PR #825, opened 2021-03-17
Validate artifact scoring

mjcarroll:
Adds the ability to score artifacts by name at their true positions using the ArtifactValidationTool.

This will only work from the ignition side, as it requires the SDF ground truth of the artifact positions.

To score an artifact, use the service call:

```
ign service -s /artifact/score \
--reqtype ignition.msgs.StringMsg \
--reptype ignition.msgs.StringMsg \
--timeout 1000 \
--req 'data: "backpack_3"'
```

The validator tool is currently still subject to the comms limitations of the environment, so distant artifacts may be "unscorable" due to radio outages.

### DIFF
--- a/subt_ign/src/ArtifactValidator.cc
+++ b/subt_ign/src/ArtifactValidator.cc
@@ -187,25 +187,28 @@ bool ArtifactValidatorPrivate::MoveToString(const std::string &_name,
 }
 
 /////////////////////////////////////////////////
-bool ArtifactValidatorPrivate::OnScore(const ignition::msgs::StringMsg& /*_req*/,
+bool ArtifactValidatorPrivate::OnScore(const ignition::msgs::StringMsg& _req,
                                        ignition::msgs::StringMsg& _rep)
 {
   // Report the artifact that we are currently at
-  
-  auto artifact = this->artifactsIter->second;
-
-  if (artifact.name == "artifact_origin")
+  //
+  auto newIter = this->artifacts.find(_req.data());
+  if (newIter == this->artifacts.end())
   {
-    std::cerr << "Attempted to score artifact_origin" << std::endl;
+    ignerr << "Artifact: " << _req.data() << " does not exist" << std::endl;
     return false;
   }
+  this->artifactsIter = newIter;
 
+  auto artifact = this->artifactsIter->second;
   std::cout << "Reporting: " << artifact.String() << std::endl; 
 
+  auto origin = this->artifacts.find("artifact_origin")->second;
+
   ignition::msgs::Pose reportPose;
-  reportPose.mutable_position()->set_x(artifact.pose.Pos().X());
-  reportPose.mutable_position()->set_y(artifact.pose.Pos().Y());
-  reportPose.mutable_position()->set_z(artifact.pose.Pos().Z());
+  reportPose.mutable_position()->set_x(artifact.pose.Pos().X() - origin.pose.Pos().X());
+  reportPose.mutable_position()->set_y(artifact.pose.Pos().Y() - origin.pose.Pos().Y());
+  reportPose.mutable_position()->set_z(artifact.pose.Pos().Z() - origin.pose.Pos().Z());
 
   return this->ReportArtifact(artifact.type,
                               reportPose);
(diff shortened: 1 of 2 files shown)


iche033:
I used this for testing the finals practice worlds. Works great. 

I just had to cycle through all the artifacts first to make sure they are loaded and test scoring artifacts after moving the validator to the `artifact_origin` so that it's in comms range.

One thing I did notice is that in finals practice world 03, the comms client had trouble binding / registering on launch:
https://github.com/osrf/subt/blob/12c844affe491744682574a0911f801f4447387b/subt_ign/src/ArtifactValidator.cc#L406

I just manually added a hack to initialize the client only after `_info.simTime` is greater than 15 seconds. I think it could be due to this world taking longer to load on my laptop.

subt_ign/src/ArtifactValidator.cc
```diff
@@ -163,6 +186,79 @@ bool ArtifactValidatorPrivate::MoveToString(const std::string &_name,
   return this->MoveTo(msg, _rep);
 }
 
+/////////////////////////////////////////////////
+bool ArtifactValidatorPrivate::OnScore(const ignition::msgs::StringMsg& _req,
+                                       ignition::msgs::StringMsg& _rep)
```
iche033, line 191: `_rep` is unused

mjcarroll:
> I just had to cycle through all the artifacts first to make sure they are loaded and test scoring artifacts after moving the validator to the artifact_origin so that it's in comms range.

Hmm, I was under the impression that this shouldn't be necessary anymore, let me double check that.

> I just manually added a hack to initialize the client only after _info.simTime is greater than 15 seconds. I think it could be due to this world taking longer to load on my laptop.

I originally had it during the constructor, which was always too early.  My understanding is that all systems should have been `Configure`-d by the time the first `PostUpdate` hits, so it shouldn't have been an issue.  I wonder if there is some other race condition in there?

iche033:
this is what I see without delaying the Bind() function:

```
[validator] CommsClient::Register: Problem registering with broker
[CommsClient] Retrying register..
[validator] CommsClient::Register: Problem registering with broker
[CommsClient] Retrying register..
[validator] CommsClient::Register: Problem registering with broker
[CommsClient] Retrying register..
[validator] CommsClient::Register: Problem registering with broker
[CommsClient] Validation service not available, invalid address or model not available
[validator] Bind() error: Trying to bind before communications are enabled!
[Msg] Moving to: artifact_origin
[Dbg] [ArtifactValidator.cc:175] Result: 1 1
[Err] [UserCommands.cc:931] Unable to update the pose for entity id:[0], name[validator]
[Dbg] [CommsBrokerPlugin.cc:218] Using [visibility_range] comms model
[Msg] Starting SubT comms broker
```

Looks like the SubT comms broker is started after the ArtifiactValidator. It could be because the CommsBroker plugin is an ign-launch plugin while the ArtifactValidator plugin is an ign-gazebo system

zwn:
> > I just had to cycle through all the artifacts first to make sure they are loaded and test scoring artifacts after moving the validator to the artifact_origin so that it's in comms range.
> 
> Hmm, I was under the impression that this shouldn't be necessary anymore, let me double check that.
> 
> > I just manually added a hack to initialize the client only after _info.simTime is greater than 15 seconds. I think it could be due to this world taking longer to load on my laptop.
> 
> I originally had it during the constructor, which was always too early. My understanding is that all systems should have been `Configure`-d by the time the first `PostUpdate` hits, so it shouldn't have been an issue. I wonder if there is some other race condition in there?

AFAIU there is no way to tell the system is "ready-to-go". Se our discussion wrt this in #412. What we ended up doing is watching "topic churn" - repeatedly listing all available ROS topics and waiting for such a time that this list is no longer growing plus a buffer of couple seconds. It's a pity something has not been designed into the system that would allow to determine with certainty that everything is done loading and is ready to go.

mjcarroll:
As far as the artifacts loading, I was mistaken, that logic remains the same in `GameLogicPlugin`.  You still have to iterate enough for all of the artifacts to be loaded in `GameLogicPlugin` (although `ArtifactValidator` is aware of all of them without concern for levels)

In terms of the sequencing issue with comms, this is purely on the ignition side without any ROS interfaces, so checking the ROS topics wouldn't reveal anything.  I think that what @iche033 said about a system plugin loaded via ign-gazebo versus a system plugin loaded via ign-plugin is probably the most likely candidate for a race condition or sequencing error.

Each ign-gazebo system should handle all configuration during the `System::Configure` step, so by the time the first `PostUpdate` step is reached, everything should be set up.

iche033 (approved):
the retry method works for me and I can now score artifacts in finals practice 03 world.

PR #844, opened 2021-03-29
Model Submission - Coordinated Robotics HD2

knoedler:
This is Coordinated Robotics version of the HD2 robot.  Same sensor platform as Karen but on the HD2 base.

Martin pointed out that their model from #842 solved our problem.  So, the problem has been resolved.  This submission had a problem.  We were unable to edit and remove the sensors from the nice .dae that team MARBLE had created.    We required assistance in figuring out how to complete that step.


peci1:
As mentioned in email, feel free to reuse our model from #842 ;)

acschang:
Continued in #1040

PR #846, opened 2021-03-29
Model Submission for Coordinated Robotics Jeanine

knoedler:
This is an upgraded version of our Jeanine robot that we used at the Urban circuit.  It is a four wheel skid steer robot using hoverboard wheels.  It uses the same sensor platform as the Allie robot.

Validation is not complete so speed and battery life are set to the defaults.

Just to vent for a moment - robots can be frustrating!  After getting plywood out on the grass to not tear up the grass again, completing the run and putting everything away - the logs were not complete.   Voltage data had dropped out just a few minutes into the battery life run.  All the other sensors kept working, just not the single one I needed.  Validation will be attempted again after the qualification deadline.


acschang (changes requested):
The initial assessment of the the model is complete however there are some issues to resolve before this model can be merged and utilized in a competitive setting:

- [x] Please add an `air_pressure` sensor to the `model.sdf` file to match the sensor list in `specifications.md`.
- [x] Please update the Arducam B0203 update rate to match the value in the specification sheet.
- [x] Please update the VLP-16's sensing range to match the value in the specification sheet.
- [ ] Please specify a minimum turning radius for the model or resolve the model's inability to rotate in place.
- [ ] Please provide validation data as it is necessary to complete the evaluation. If not provided, the model's maximum velocity will be limited to 1m/s and battery life set to 60min. When the model validation data can be provided later, please submit a PR updating the model.

Please address the issues noted above in this pull request. Once the above issues are fixed, your submission will move onto the next review stage!

acschang (approved):

submitted_models/coro_jeanine_sensor_config_1/launch/vehicle_topics.launch
```diff
@@ -0,0 +1,276 @@
+<?xml version="1.0"?>
+<launch>
```
iche033, line 2: the `IGN_TRANSPORT_TOPIC_STATISTICS` env var is msising:

```
<env name="IGN_TRANSPORT_TOPIC_STATISTICS" value="1"/>
```

from step 8: https://github.com/osrf/subt/wiki/Model%20Submission%20Process#adding-a-custom-configuration-to-existing-vehicles

iche033 (commented):
sensor data looks fine in rviz, optical frame works, tf tree looks good, vehicle moves as expected, battery params are good, and config 2's breadcrumb deployment works fine,

Give then the spec says it can carry a drone on its sensor platform, we'll need to add it to the masupial vehicle parent list later.

I have one comment about adding the `IGN_TRANSPORT_TOPIC_STATISTICS`. Apart from that this looks good to me.

submitted_models/coro_jeanine_sensor_config_1/launch/vehicle_topics.launch
nkoenig: I'll make this change in a follow up pr with marsupials pairs.

PR #857, opened 2021-03-30
Robotika Kloubak config 1 - fix cameras, lidar and symmetry

m3d:
The wide angle cameras are not properly simulated so identical version to Freyja is used now.
URFD was regenerated to use proper meshes, fixed position of lidars and symmetry of robot.


m3d:
these errors were highlighted during preparation of config 3 (#855) and fixes are already used there too

m3d:
I checked the release from 2021-04-19 and I can confirm this fix is still not there yet ;)
saveX-0000
It is pity because Kloubak is one of few wheeled robots capable to handle the rails - any estimate when this could be patched?

m3d:
@nkoenig can you please at least update model on FUEL from git (I mean current master)? (for local model True at least the lights work, according to my colleagues)

PR #858, opened 2021-03-30
KineticEnergy and HaltMotion added to models

Lobotuerk:
Signed-off-by: Tomas Lorente <jtlorente@ekumenlabs.com>
Added KineticEnergy Threshold surpassed event logging to GameLogic. Added the kineticEnergyMonitor to X1, X2, X3 and X4

Depends on https://github.com/ignitionrobotics/ign-gazebo/pull/728

Testing: Change `<halt_mode>false</halt_mode>` to `<halt_mode>true</halt_mode>` for the robot configuration you want to test   haltMotion . Threshold set to default 100, but can be changed by modifying `<kinetic_energy_threshold>100</kinetic_energy_threshold>`

### DIFF
--- a/subt_ign/src/GameLogicPlugin.cc
+++ b/subt_ign/src/GameLogicPlugin.cc
@@ -1067,7 +1067,7 @@ void GameLogicPlugin::PreUpdate(const UpdateInfo &_info,
     {
       if (this->dataPtr->robotEntityMap.find(robotName) ==
           this->dataPtr->robotEntityMap.end())
-          return;
+          continue;
       gazebo::Entity robotEntity = this->dataPtr->robotEntityMap.at(robotName);
       auto *haltMotionComp = _ecm.Component<components::HaltMotion>(robotEntity);
       if (haltMotionComp && !haltMotionComp->Data())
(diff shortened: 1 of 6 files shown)


subt_ign/src/GameLogicPlugin.cc
```diff
@@ -2748,3 +2762,33 @@ double GameLogicPluginPrivate::FloorMultiple(double _n, double _m)
...
+  if (topicParts.size() > 1)
+    name = topicParts[1];
+
+  {
+    std::lock_guard<std::mutex> lock(this->eventCounterMutex);
+    std::ostringstream stream;
+    stream
+      << "- event:\n"
+      << "  id: " << this->eventCounter << "\n"
+      << "  type: KineticEnergyThreshold Surpassed\n"
```
angelacmaio, line 2786: Recommend renaming the event type 
```suggestion
      << "  type: collision\n"
```

subt_ign/src/GameLogicPlugin.cc
```diff
@@ -1012,6 +1024,60 @@ void GameLogicPluginPrivate::OnEvent(const ignition::msgs::Pose &_msg)
...
+          return true;
+        });
+  }
+  else
+  {
+    for (std::string robotName : this->dataPtr->haltedRobots)
+    {
+      if (this->dataPtr->robotEntityMap.find(robotName) ==
+          this->dataPtr->robotEntityMap.end())
+          return;
```
nkoenig, line 1070:
```suggestion
          continue;
```

Lobotuerk: Agree, but I think a case could be made about why do you have a halted robot that is not inside the Entity map

PR #859, opened 2021-03-31
Model submission for Absolem sensor configurations 3,4,5 by CTU-CRAS-Norlab

peci1:
Depends and builds on top of #877 and #878. If you want to see just changes from the sensor configurations, look e.g. at https://github.com/ctu-vras/subt/compare/submitted_models/ctu_cras_norlab_absolem/remove_annoying_visuals...ctu-vras:submitted_models/ctu_cras_norlab_absolem_improved .

The Absolem robots are just now undergoing a facelift which removes the tilting 2D lidar, adds a 3D lidar and makes necessary changes to the shape and weight distribution of the body. Additionally, we have identified and found a way to fix a source of very strong vibrations when the robot was driving 0.5 - 0.6 m/s. It is mechanically capable of driving using this velocity, but the vibrations made the sensor data unusable. After fixing the vibrations, it is okay to drive 0.5 or maybe even 0.6 m/s. So I increased the velocity limit for sensor configs 3, 4 and 5 to 0.6 m/s (specified in `sensor_config_3/launch/common.rb`). I didn't touch the weight of the body as it should be more or less the same as before (we remove a 2.5 kg lidar and exchange it for a 0.5 kg one, but we extend the body a bit, so it should cancel out). What will change is weight distribution and now the center of gravity will be much closer to the track hinges.

The facelift has deadline in 7 weeks, so if any validation data with the real robot are required, we will provide them later. All sensors are the same as on our MARV robot, so I suggest assuming they do not need explicit validation in this PR. I managed to fix some of the problems @adlarkin discovered on MARV.

This PR touches spawner.rb and vehicle_topics.launch of sensor config 1 quite heavily, but all changes are in the name of better maintainability and "extensibility" by other sensor configs. I explicitly verified that the changes did not break anything on sensor configs 1 and 2 (all data are coming to ROS, I can drive, I can control flippers, I can drop breadcrumbs). The changes to IMU and wheel slip made by SubT organizers were retained and verified to still do the same thing. The only thing that has changed for sensor configs 1 and 2 is the rate at which odometry is published. As discovered in the MARV model, each of the 5 diffdrive plugins published 50 Hz odometry to the same topic, so there were 250 messages per simulated second. I removed the 4 superfluous publishers and left only the one publishing odometry of the main tracks, so now the odometry topic has rate of 50 Hz.

PR #551, opened 2020-08-17
New features for Absolem robot.

peci1:
This PR depends on #542.

New features were added to the Absolem robot model:

- relative positional control of flippers
- velocity and torque limits from `<joint><limits>` are now respected by the flipper controller
- the FlipperControlPlugin was rewritten to use a more reliable implementation of the switching between positional and velocity control

PR #860, opened 2021-03-31
Model submission for Absolem sensor configurations 6,7,8 by CTU-CRAS-Norlab

peci1:
Depends and builds on top of #859 (sensor configs 3,4,5), #877 and #878. That is why you see so much commits. If you want to see just changes from the sensor configurations, look e.g. at https://github.com/ctu-vras/subt/compare/submitted_models/ctu_cras_norlab_absolem_improved...ctu-vras:submitted_models/ctu_cras_norlab_absolem/clone .

Everything that was written in #859 applies also here. The main difference of these sensor configs is that they do not have the Ladybug omnicamera, and instead they use our custom omnicamera solution based on 5 pcs of Basler ACE 2 Pro cameras. This is beneficial for occlusions by body parts, which are inevitable when using a centralized omnicam, but are nicely avoided by putting each camera on the edge of the robot body.

The field of view of the Basler cameras is experimentally measured. We took a ruler, shot a picture of it, and by using a bit of trigonometry, we ended up with horizontal FOV of 86.5°. For the trigonometrical computation we used the datasheet value of "Front focal length", which is 9.95 mm. We sketched a triangle from the point on lens axis that is 9.95 mm inside the lens (measured from the vertex of the outern-most lens). One side of the triangle we used for computation is distance of the ruler from the point inside lens, and the other side is the distance viewed on the ruler.

The theoretical value for a 4 mm lens on 1/2.3" chip is 79°, so I assume the difference to our measured value is done by distortion, which is not modeled in the simulator, and neither is it modeled in the theoretical equation.

The computation I used for horizontal FOV is:

    degrees(2*atan(tan(0.0066)/(2*0.004))) = 79.04 

Where `0.004` is effective focal length and `0.0066` is the sensor width (both from datasheet). This theoretical value is valid for focusing into infinity. Closer focus can decrease the FOV by a few degrees.


peci1:
I force-pushed a SDF 1.6 version of Absolem SC6-8 based on #877 and #878. The geometry and behavior is exactly the same as in the original version.

peci1:
I rebased on top of #859 to resolve conflicts brought in by #551.

acschang (changes requested):
The initial assessment of these additional sensor configurations for the Absolem model is complete, however there are some issues to resolve before these additional sensor configurations can be merged and utilized in a competitive setting:

- [x] Please add lens intrinsic values to the Omnicam cameras.
- [x] Please add all missing components to the RViz visualization.
- [x] Please include all updates from Absolem PR for sensor configurations 3, 4, and 5 (#859).
- [x] Please regenerate the `model.sdf` for sensor configurations 6, 7, and 8 with the updates from #859.

Please address the issues noted above and add commits to this pull request as soon as possible. The submission is expected to be incorporated into the SubT Virtual Testbed pending a successful review.

peci1:
All of the issues from last review have been resolved.

acschang (approved):

nkoenig (approved):

PR #864, opened 2021-03-31
Pull Request for Coordinated Robotics' Mark UAV

knoedler:
This is the Mark UAV that was used at the tunnel circuit.  Validation tests have not been completed so the velocity is limited to 4m/S and battery life to 10 minutes.


acschang:
Continued in #1039

PR #872, opened 2021-04-01
 Model submission for Balloon sensor config 1 by CTU-CRAS-Norlab 

peci1:
This PR adds a helium balloon that can be attached to UGVs to help them explore high ceilings. The balloon is on a reel so that it can be extended and retracted as needed. It is added to UGVs as a marsupial child, but it should never be detached (but I haven't found a way to forbid that).

This PR also adds the Buoyancy system to competition.ign with air density (1.27 kg/m3). This should not cause any problems to any models with considerable weight. But if someone forgot to specify weight for something, it will fly to the skies. If adding this system is a blocker, I can exchange it for a model-specific plugin that would apply a constant force along gravity. But I though using the buoyancy plugin is the more correct approach.

Until I tuned the joints, I saw a few crashes in DART/ODE who told there were NaNs in LCP solution. However, after adding friction to the joints, I've never seen these segfaults. But it is good to keep an eye on them as I'm not really sure what was causing them, and the friction just probably helped keeping all numbers reasonably low.


peci1:
https://user-images.githubusercontent.com/182533/113280498-aa29d880-92e4-11eb-8daa-6f57eb68a29d.mp4

peci1:
For easier testing, I added a republish from one of the flipper topics controlled by gamepad from subt_example. Don't forget the balloon is controlled from itself, so you have to choose it on the color buttons, and then d-pad up-down.

peci1:

angelacmaio:
This is an interesting and unique feature! We would be happy to add, but unfortunately the implementation appears to differ too substantively from the real balloon to be validated in accordance with the Model Preparation Guide. 

Implemented as a separate platform (i.e., as a marsupial child), there are discrepancies compared to the real balloon which:

- does not have onboard computing capability
- cannot detach from the UGV
- does not have communications with other robots except the attached
- presumably, is able to send full sensor streams to the UGV (vs. being limited to 1.5KB packets by the comms model)
- presumably, does not control its own tether length

With an abundance of time, we would recommend alternate implementations, e.g., to model the balloon as an extendable sensor configuration for UGVs, or to work with another group to model an actuated blimp and gather validation data from the actual platform.

We do appreciate the unique submission and recommend uploading this model to [Ignition Fuel](https://app.ignitionrobotics.org/fuel/models) (see instructions [here](https://ignitionrobotics.org/api/gazebo/3.2/meshtofuel.html)) and linking to your forked repository in the model description so that others may test with it outside of the competition context.

peci1:
> With an abundance of time, we would recommend alternate implementations

Forgive my imperfect English understanding. So do you actually suggest us to provide an alternative implementation or not? How would that play with the model submission deadline?

angelacmaio:
> Forgive my imperfect English understanding. So do you actually suggest us to provide an alternative implementation or not? How would that play with the model submission deadline?

The model submission deadline has passed, so we will not be able to incorporate alternative implementations of this model into the SubT Tech Repo for the Final Event. As a reminder, you can still make the model publicly accessible on [Ignition Fuel](https://app.ignitionrobotics.org/fuel/models) if you wish. We appreciate and look forward to the continued community engagement!

PR #852, opened 2021-03-30
Absolem: Remove debug visuals

peci1:
There are some visuals we forgotten to remove on Absolem. They create annoying artifacts in the omnicamera. This PR removes the visuals (all of them were black cubes 0.5x0.5x0.5 cm I used to debug placement of the camera sensors).

Cameras without this PR:
absolem_unfixed

Cameras with this PR:
absolem_fixed

I based this PR on top of #851 . After merging #851, this will be just a single commit. Look at the last commit in this PR during review only.

PR #875, opened 2021-04-01
Absolem visuals fix

nkoenig:
Follow on to #852 where I change SDF to use version 1.6.

### DIFF
--- a/submitted_models/ctu_cras_norlab_absolem_sensor_config_1/model.sdf
+++ b/submitted_models/ctu_cras_norlab_absolem_sensor_config_1/model.sdf
@@ -1,5 +1,5 @@
 <?xml version='1.0' encoding='utf-8'?>
-<sdf version="1.6">
+<sdf version="1.7">
   <model name="ctu_cras_norlab_absolem_sensor_config_1">
     <link name="base_link">
       <inertial>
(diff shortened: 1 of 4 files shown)


peci1:
Thank you for the PR. I'll review it over the weekend. What is the approach you used? `ign sdf -p` in blueprint version of sdformat, or manual editing approach?

Also, did you see any other SDF 1.7-only feature used in our SDFs than `relative_to` poses?

peci1:
I reviewed this PR and it produces invalid Ignition SDF. It e.g. contains the `model://ctu_cras_norlab_absolem_sensor_config_1/` prefix for meshes which is unresolvable in Ignition Gazebo. It also did not use the Xacro parameters of the robot that are set in `config/common.sh` and `config/sdf.sh`, which results in a somewhat different robot (it e.g. has the differential connecting tracks to body, which creates a closed kinematic loop, and those are not representable in IG4/DART).

These are all things our `update_robot_sdf` script takes care about to be set correctly.

I just installed `libsdformat8-dev` and forced calling `ign sdf --force-version 8.9.1 -p` in the `update_robot_sdf` script. That results in a very minimal change to the `model.sdf` present on master branch (keeping SDF version 1.6), while allowing to continually develop the robot model for IG4.

I suggest closing this PR and I'll update all of our robot model PRs to generate the SDF using libsdformat 8. Using this appraoch, users of the model will not be affected in any way. Only developers who want to update the robot model will need to install `libsdformat8-dev` package.

PR #883, opened 2021-04-07
Submitted models/ctu cras norlab x500 sensor config 2

petrlmat:
Sensor config 2 of ctu_cras_norlab_x500 UAV is the same as sensor config 1 without up and down Realsense cameras. It consists of the following sensors:
- 3D medium range LiDAR - Ouster OS0-64
- HD wide camera - Basler dart daA1600-60um with DSL165A lens
- IMU with standard SubT specification


angelacmaio:
This model will not be incorporated into the Virtual Competition, as it was submitted after the March 31st Robot Model Submission Deadline. We would recommend uploading the model to [Fuel](https://app.ignitionrobotics.org/fuel/models) so it may be utilized outside of the competition. Thank you for the contribution.

PR #901, opened 2021-04-21
Move battery plugin to the end of plugin list

mjcarroll:
Uncovered by @azeey  in https://github.com/osrf/subt/pull/868#pullrequestreview-638036318

This should not require an update in Fuel, as it purely impacts the launch/spawning of the model.

cc: @peci1


peci1 (approved):

peci1:
Oh, good catch. What is the reason for this behavior?

mjcarroll:
I believe that it is based on the order of instantiation of the plugins.  This guarantees that all joint entities are created before the battery plugin in loaded.

nkoenig (approved):

PR #903, opened 2021-04-22
Fixes pose of rs_up, rs_down rgbd cameras of x500 UAV

petrlmat:
There was a missing zero in the sensor pose.


knoedler:
Should the z on the down camera be -0.2 or should it be -0.02?

nkoenig (approved):

PR #908, opened 2021-04-28
Model Submission for Marble Spot

acschang:
Continuation of #868 

**Adjustments:**
- Adjusted references from CTU-CRAS-Norlab Spot to Marble Spot sensor config 1+2 where possible.
- Fixed the resolution and range of the OS1.
- Included and updated IMU parameters to match the standard for SubT.
- Added base model updates from the `BOSDYN_SPOT` model.


mjcarroll:
Can we close #868 since this replaces it?

nkoenig (approved):

peci1:
It seems the thumbnails folder in sensor config 1 actually contains thumbnail of CTU-CRAS-Norlab Spot...

PR #779, opened 2021-02-03
New config with thermal camera: MARBLE_HUSKY_SENSOR_CONFIG_3 and MARBLE_HUSKY_SENSOR_CONFIG_4

adlarkin:
The thermal camera used for this model is the same as the one in #773. The only difference between the thermal camera for this model and the one in #773 should be the the camera's location.

`sensor config 3 = sensor config 1 + thermal camera`, and `sensor config 4 = sensor config 2 + thermal camera`

Here is where I have placed the thermal camera (right above the lower LIDAR sensor): 
husky_thermal_camera

Signed-off-by: Ashton Larkin <ashton@openrobotics.org>

PR #777, opened 2021-02-03
New config with thermal camera: CERBERUS_M100_SENSOR_CONFIG_2

iche033:
Created a new config: `CERBERUS_M100_SENSOR_CONFIG_2`: 

This is a copy of `CERBERUS_M100_SENSOR_CONFIG_1` with thermal camera sensor added to `model.sdf` and `vehicle_topics.launch` updated to bridge the thermal camera topics

Still need clarification on (**Edit: done**):
* ~~exact FLIR Tau2 model used (there are a few product variations on the [FLIR Tau2 page](https://www.flir.com/products/tau-2/))~~
    * Modeled after [this configuration](https://www.flir.com/products/tau-2/?model=46640007H)
* ~~location of the sensor on the drone~~
    * Adjusted camera and thermal camera pose

cc @tiralonghipol @mihirk284 

The thermal camera is mounted side by side with the rgb camera:

cerberus_m100_config_2_thermal

to test:

```
ign launch cave_circuit.ign worldName:=simple_cave_01 circuit:=cave robotName1:=Xa robotConfig1:=CERBERUS_M100_SENSOR_CONFIG_2
```

Signed-off-by: Ian Chen <ichen@osrfoundation.org>

Issue #406, opened 2020-04-21
Camera image sensor center (principal point) off by 1.


pauljurczak:
It seems that most (all) camera principal point coordinates in `model.sdf` files and consequently in `sensor_msgs/CameraInfo` are off by 1. For example, 640x480 VGA RGBD camera has principal point set to (320.5, 240.5). Assuming the lowest pixel coordinates of (0, 0), principal point (geometric center of image sensor) of this camera is (319.5, 239.5). It is given by `(width/2-0.5, height/2-0.5)` equation not `((width+1)/2, (height+1)/2)` used in `model.sdf` file.

Theoretically, the principal point can be placed anywhere on image sensor plane, but I'm assuming that your intention was to place it centrally to achieve a symmetric pinhole camera model.

acschang:
The discrepancy in the principle point is reflected in the `camera_info` topic for each camera stream.  Fixing the reported issue would require a subtle change to the majority of cameras modeled in the repository which could adversely affect team solutions thus this issue cannot be resolved until after Finals.

pauljurczak:
No problem. This is very minor, so I'm closing it now.

PR #910, opened 2021-04-30
Thermal Camera Consistency

acschang:
Updated the thermal cameras on MARV, Mike, and Karen to meet the standards established in #773, #776, #777, and #779.

**MARV:**
marv_thermal

**Mike:**
mike_thermal

**Karen:**
karen_thermal


peci1:
Please, do not decrease near clip distance on MARV. The value that is there is not arbitrary - it is a result of extensive testing. As mentioned in some of our models earlier, Gazebo has some rendering issues when the robot changes orientation. Not only yaw, but also pitch/roll play a role. If the near clip distance is too low, the thermocamera image gets blinded by the camera mesh in some situations.

peci1:
It seems there is inconsistency in the camera center computation (along the lines pointed in https://github.com/osrf/subt/issues/406 ). Some robots got -1 here and some got +1. Can you please definitively decide what is the correct formula?

acschang:
> Please, do not decrease near clip distance on MARV. The value that is there is not arbitrary - it is a result of extensive testing. As mentioned in some of our models earlier, Gazebo has some rendering issues when the robot changes orientation. Not only yaw, but also pitch/roll play a role. If the near clip distance is too low, the thermocamera image gets blinded by the camera mesh in some situations.

Noted.  The clip distance is not expected to be adjusted with this PR.

> It seems there is inconsistency in the camera center computation (along the lines pointed in #406 ). Some robots got -1 here and some got +1. Can you please definitively decide what is the correct formula?

The standard is +1 as noted in the following:
```
<!-- fx = fy = width / ( 2 * tan (hfov / 2 ) ) -->
<!-- cx = ( width + 1 ) / 2 -->
<!-- cy = ( height + 1 ) / 2 -->
```

I'll look into resolving the issue pointed out in #406.

mjcarroll:
I believe this was still waiting on @acschang resolve the issues from #406.  There are still a few cameras in here with inconsistent updates to the principle points.  I can make the changes if need be.

acschang:
> > Please, do not decrease near clip distance on MARV. The value that is there is not arbitrary - it is a result of extensive testing. As mentioned in some of our models earlier, Gazebo has some rendering issues when the robot changes orientation. Not only yaw, but also pitch/roll play a role. If the near clip distance is too low, the thermocamera image gets blinded by the camera mesh in some situations.
> 
> Noted. The clip distance is not expected to be adjusted with this PR.
> 
> > It seems there is inconsistency in the camera center computation (along the lines pointed in #406 ). Some robots got -1 here and some got +1. Can you please definitively decide what is the correct formula?
> 
> The standard is +1 as noted in the following:
> 
> ```
> 
> 
> 
> ```
> 
> I'll look into resolving the issue pointed out in #406.

It appears that the #406 is correct and the values for cx/cy are off by 1 pixel for a number of models but they will not adjusted for finals as the inconsistency for the principle point can be derived from `camera_info` and correcting the issue will change behavior that has been present since the transition to Ignition.

However, I did adjust the thermal camera principle point for MARV to align with the principle point in other camera sensors for this vehicle model.

acschang:
> I believe this was still waiting on @acschang resolve the issues from #406. There are still a few cameras in here with inconsistent updates to the principle points. I can make the changes if need be.

This PR is good to go.  We are not going to adjust camera cx/cy values in other models before finals.

peci1:
Should this be retargeted to final_event branch?

mjcarroll (approved):

PR #916, opened 2021-05-17
Implements vehicle damage based on kinetic energy

nkoenig:
This is an alternative to PR #858. I'm using the kinetic energy formula, `KE = 0.5 * M * v * v`, to calculate a kinetic energy threshold. If the change in KE between iterations passes this threshold, then a crash has occurred and the robot will be disabled. The value of `v` is set to the velocity achieved by falling a certain distance. Currently this distance is 0.077m.

Requires: https://github.com/ignitionrobotics/ign-gazebo/pull/728


Lobotuerk (approved):

subt_ign/src/GameLogicPlugin.cc
```diff
@@ -1012,6 +1040,163 @@ void GameLogicPluginPrivate::OnEvent(const ignition::msgs::Pose &_msg)
...
+                 _ecm.CreateComponent(model->Data(), components::HaltMotion());
+              }
+            }
+          }
+          return true;
+        });
+  }
+
+  // Check for crashes
+  for (auto &ke : this->dataPtr->keInfo)
```
Lobotuerk, line 1154: Should not this part be inside an `else` after `(!this->dataPtr->started)` to make sure no crash happens before starting, like on spawn?

subt_ign/src/GameLogicPlugin.cc
```diff
@@ -1012,6 +1049,163 @@ void GameLogicPluginPrivate::OnEvent(const ignition::msgs::Pose &_msg)
...
+          localSimTime.set_sec(sec);
+          localSimTime.set_nsec(nsec);
+
+          std::lock_guard<std::mutex> lock(this->dataPtr->eventCounterMutex);
+          // _resp.set_report_status("scoring finished");
+          std::ostringstream stream;
+          stream
+            << "- event:\n"
+            << "  id: " << this->dataPtr->eventCounter << "\n"
+            << "  type: kinetic\n"
```
angelacmaio, line 1188: Can we use "collision" as the event name instead?

subt_ign/src/GameLogicPlugin.cc
```diff
@@ -1012,6 +1049,163 @@ void GameLogicPluginPrivate::OnEvent(const ignition::msgs::Pose &_msg)
...
+        {
+          int64_t sec, nsec;
+          std::tie(sec, nsec) =
+            ignition::math::durationToSecNsec(_info.simTime);
+          ignition::msgs::Time localSimTime;
+          localSimTime.set_sec(sec);
+          localSimTime.set_nsec(nsec);
+
+          std::lock_guard<std::mutex> lock(this->dataPtr->eventCounterMutex);
+          // _resp.set_report_status("scoring finished");
```
angelacmaio, line 1183: Comment can be removed

peci1:
#920 inadvarently merged this

nkoenig:
> #920 inadvarently merged this

Thanks. Reverted in d83b71b.

subt_ign/src/GameLogicPlugin.cc
nkoenig: Done in 56f62bcb12661f5e8338fa7f9abdadaa22775b59

subt_ign/src/GameLogicPlugin.cc
nkoenig: Removed in 56f62bcb12661f5e8338fa7f9abdadaa22775b59

subt_ign/src/GameLogicPlugin.cc
```diff
@@ -1012,6 +1049,174 @@ void GameLogicPluginPrivate::OnEvent(const ignition::msgs::Pose &_msg)
...
+            {
+              // This sets the kinetic energy threshold for the robot. It is
+              // based on the kinetic energy formula KE = 0.5 * M * v * v
+              //
+              // M is the mass of the robot.
+              // v is velocity of the robot. We are using the velocity acheived
+              // by falling from a height.
+              this->dataPtr->keInfo[model->Data()].kineticEnergyThreshold =
+                0.5 * mass * std::pow(
+                    sqrt((2 * this->dataPtr->keHeight) / 9.8) * 9.8, 2);
```
acschang, line 1143: This ends up evaluating equivocally however the velocity can be simplified from:
`sqrt((2 * this->dataPtr->keHeight) / 9.8) * 9.8`
to:
`sqrt((2 * this->dataPtr->keHeight) * 9.8)`
which would omit the cost of one division.

nkoenig: Done in 102aa622dc0ad1b5eeef35903dc42774ed889f7c

subt_ign/include/subt_ign/RobotPlatformTypes.hh
```diff
@@ -23,36 +23,36 @@ const std::map<std::string, double> robotPlatformTypes = {
...
+  {"SHAFTER", 8}, // UAV, no prop guards
   {"SPOT", 1},
   {"TEAMBASE", 1},
   {"X1", 1},
   {"X2", 1},
-  {"X3", 7.2},
-  {"X4", 7.2},
-  {"X500", 14}
+  {"X3", 8}, // UAV, no prop guards
+  {"X6", 8}, // UAV, no prop guards
```
acschang, line 54: @nkoenig was changing this from `X4` to `X6` intended?

acschang: Fixed in https://github.com/osrf/subt/pull/916/commits/3a7692b8615143a021251ee2efe1ee2fc15b897d

PR #919, opened 2021-05-18
Make gzserver and bridge images log stdout into a log file.

marcoshuck:
### Context
The cloudsim team would like add a new feature to upload logs and enable different teams to download robot log messages from the standard output. This will also help Ignition Gazebo developers use those logs to debug errors.

### Change
This PR appends a `tee` command to the `gzserver` and `bridge` entry-point scripts in order to generate a log file that will be collected by cloudsim and uploaded to a storage for future use.

### DIFF
--- a/docker/cloudsim_bridge/run_bridge.bash
+++ b/docker/cloudsim_bridge/run_bridge.bash
@@ -21,4 +21,4 @@ export ROSCONSOLE_STDOUT_LINE_BUFFERED=1
 # atop process monitoring
 atop -R -w /home/developer/.ros/atop_log &
 
-ign launch cloudsim_bridge.ign -v 4 $@
+ign launch cloudsim_bridge.ign -v 4 $@ | tee /home/developer/.ros/bridge.log
--- a/docker/cloudsim_bridge/run_bridge.bash
+++ b/docker/cloudsim_bridge/run_bridge.bash
@@ -21,4 +21,4 @@ export ROSCONSOLE_STDOUT_LINE_BUFFERED=1
 # atop process monitoring
 atop -R -w /home/developer/.ros/atop_log &
 
-ign launch cloudsim_bridge.ign -v 4 "$@" | tee /home/developer/.ros/bridge.log
+ign launch cloudsim_bridge.ign -v 4 "$@" 2>&1 | tee /home/developer/.ros/bridge.log
--- a/docker/cloudsim_bridge/run_bridge.bash
+++ b/docker/cloudsim_bridge/run_bridge.bash
@@ -21,4 +21,4 @@ export ROSCONSOLE_STDOUT_LINE_BUFFERED=1
 # atop process monitoring
 atop -R -w /home/developer/.ros/atop_log &
 
-ign launch cloudsim_bridge.ign -v 4 $@ | tee /home/developer/.ros/bridge.log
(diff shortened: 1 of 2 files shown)


docker/cloudsim_bridge/run_bridge.bash
```diff
@@ -21,4 +21,4 @@ export ROSCONSOLE_STDOUT_LINE_BUFFERED=1
 # atop process monitoring
 atop -R -w /home/developer/.ros/atop_log &
 
-ign launch cloudsim_bridge.ign -v 4 $@
+ign launch cloudsim_bridge.ign -v 4 $@ | tee /home/developer/.ros/bridge.log
```
AlejoAsd, line 24: We should pipe `stderr` as well.

```suggestion
ign launch cloudsim_bridge.ign -v 4 $@ |& tee /home/developer/.ros/bridge.log
```

docker/cloudsim_sim/run_sim.bash
```diff
@@ -14,4 +14,4 @@ bwm-ng -o csv -c 0 -t 1000 -T rate -I eth0 >> $FILE &
 atop -R -w /tmp/ign/logs/atop_log &
 
 export ROS_LOG_DIR=/tmp/ign/logs/ros
-ign launch -v 4 $@
+ign launch -v 4 $@ | tee /tmp/ign/logs/gzserver.log
```
AlejoAsd, line 17: We should pipe `stderr` as well.

```suggestion
ign launch -v 4 $@ |& tee /tmp/ign/logs/gzserver.log
```

AlejoAsd (changes requested):

docker/cloudsim_bridge/run_bridge.bash
marcoshuck: Do we want to log `stderr` in the `stdout` too?

docker/cloudsim_sim/run_sim.bash
marcoshuck: Do we want to log `stderr` in the `stdout` too?

docker/cloudsim_bridge/run_bridge.bash
AlejoAsd: Yes, we will lose that information if we don't.

marcoshuck: We can copy whatever comes from stderr into stdout and then logging that into a file by doing this:
`ign launch ... 2>&1 | tee`

nkoenig:
Did this work for you? When I run it locally, not inside of docker, I don't get a log file in /tmp/ign/logs.

iche033:
the file is probably created but then removed by ign-launch when it removes and recreates the `/tmp/ign/logs` log dir:
https://github.com/ignitionrobotics/ign-launch/blob/ign-launch4/plugins/gazebo_server/GazeboServer.cc#L149-L154

seems to work if I specify a different path

marcoshuck:
> the file is probably created but then removed by ign-launch when it removes and recreates the `/tmp/ign/logs` log dir:
> https://github.com/ignitionrobotics/ign-launch/blob/ign-launch4/plugins/gazebo_server/GazeboServer.cc#L149-L154
> 
> seems to work if I specify a different path

@nkoenig exactly what Ian mentioned, I managed to sort out this problem by setting the `logPath` argument when using the `cloudsim_sim.ign` launch file.

PR #926, opened 2021-05-20
Adding crash factor to robot types

nkoenig:
Adds support for a KE factor per robot type. There is a debug statement in the code right now that outputs the factor required to cause a disabling collision. This can be used to help determine good values for each platform.

Note: Collisions are only enabled once state is `started`.


nkoenig:
Merged this into PR #916.

PR #929, opened 2021-05-24
Fix X500 RViz model

peci1:
There were incorrect links to meshes for X500 in URDF. They are fixed now.

Also, the relative pose of the props had to be changed, as this is how it looked in RViz:

https://user-images.githubusercontent.com/182533/119370610-c2e7b680-bcb5-11eb-94b8-aa33e9a3a581.mp4

This PR fixes the props (both visuals and collisions) to proper positions.


peci1:
CC @petrlmat

peci1:
The video doesn't play in Firefox/Linux for some reason, but Chromium plays it okay

mjcarroll (approved):

PR #920, opened 2021-05-18
Model Submission for MARBLE Husky SC6

acschang:
Continuation of #850 and #918.  Both of these PRs have a conflicting Husky SC3 from #517 that was never merged and then displaced by #779.

PR #941, opened 2021-05-28
Marsupial update  for Marble Husky

bfotheri:
As requested [here](https://github.com/osrf/subt/pull/920#issuecomment-848776550****), the Marble Husky Vehicles have been updated for marsupial ability. Earlier sensor configurations were also marsupial compatible (in hardware) but the capability did not exist yet in simulation at the time of their making. Let me know if we need any other data or changes for this PR. Thanks guys!

### DIFF
--- a/subt_ign/launch/competition.ign
+++ b/subt_ign/launch/competition.ign
@@ -116,6 +116,7 @@
 
   MARSUPIAL_PARENT_LINK_NAMES = {
      "X1" => "base_link",
+     "MARBLE_HUSKY" => "base_link",
      "EXPLORER_X1" => "base_link",
      "EXPLORER_R2" => "Rear_Rocker_Link",
      "CSIRO_DATA61_OZBOT_ATR" => "base_link",
@@ -127,6 +128,7 @@
 
   MARSUPIAL_VALID_ROBOT_PAIRS = {
     "X1" => ["X3", "MARBLE_QAV500", "CERBERUS_GAGARIN", "CERBERUS_RMF", "COSTAR_SHAFTER", "CORO_PAM", "CTU_CRAS_NORLAB_X500", "CORO_CRYSTAL"],
+    "MARBLE_HUSKY" =>["X3", "MARBLE_QAV500", "CERBERUS_GAGARIN", "CERBERUS_RMF", "COSTAR_SHAFTER", "CORO_PAM", "CTU_CRAS_NORLAB_X500", "CORO_CRYSTAL"],
     "CORO_ALLIE" => ["CERBERUS_RMF", "CORO_PAM", "CORO_CRYSTAL"],
     "CORO_JEANINE" => ["CERBERUS_RMF", "CORO_PAM", "CORO_CRYSTAL"],
     "CORO_KAREN" => ["CERBERUS_RMF", "CORO_PAM", "CORO_CRYSTAL"],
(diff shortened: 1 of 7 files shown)


subt_ign/launch/competition.ign
```diff
@@ -300,7 +304,7 @@
     <record>
       <enabled>true</enabled>
       <!-- This path is used by cloudsim, please do not change -->
-      <path>/tmp/ign/logs/gazebo</path>
+      <path>/tmp/ign/logs</path>
```
iche033, line 307: can you revert this change and the changes below to `static_update_frequency`?

subt_ign/launch/competition.ign
```diff
@@ -138,6 +140,7 @@
 
   MARSUPIAL_PARENT_ROBOT_POSITION_OFFSETS = {
     "X1" => Vector3d.new(0, 0, 0.717),
+    "MARBLE_HUSKY" => Vector3d.new(-0.06, 0, 0.392),
```
iche033, line 143: looks like the offset needs to be higher. The UAVs are currently overlapping with the landing pad. Most obvious when testing with X3:

marble_husky_x3_marsupial

angelacmaio: @bfotheri pinging on the above feedback

subt_ign/launch/competition.ign
dan-riley: The static_update_frequency changes really need to be made, in this or another PR.  Ground truth is broken unless those are increased.

subt_ign/launch/competition.ign
dan-riley: The offsets were based on the QAV500.  Also, it looks like your visualization of the Husky didn't use the updated sdf?  The black platform should look different.

bfotheri: I believe the black platform is updated for sensor config 5 but that is how it looked originally for the computing block.

iche033: I think the model on Fuel may not have updated. I can fix that. I'm now testing with `localModel:=true`. This is what I see:

husky_c5_marsupial

Is this the updated mesh? There is still a light overlap between the two.

subt_ign/launch/competition.ign
iche033: let's try get the husky changes in first in this PR and review the other launch file changes separately.

subt_ign/launch/competition.ign
bfotheri: That does appear to be the updated mesh. I'll push a commit to fix that offset and the static_update_frequency param

bfotheri: Commit pushed.

subt_ign/launch/competition.ign
```diff
@@ -367,7 +371,6 @@
       <!-- The collection of artifacts to locate -->
       <world_name><%= $worldName %></world_name>
       <ros>false</ros>
-      <ke_height>0.078</ke_height>
```
bfotheri, line 370: Was this line added as an update? I don't remember removing it but may have accidentally. Does it need to be reverted as well?

bfotheri:
Are there any more items that need to be addressed before merging this?

dan-riley:
If this could get approved in enough time that we could use this for the Virtual final submission we'd appreciate it.  Thanks!

Issue #945, opened 2021-06-02
Server output is now buffered by large pieces


peci1:
With  #919, the server log started to be buffered by pretty large pieces (half screen of text). See explanation here: https://stackoverflow.com/questions/11337041/force-line-buffering-of-stdout-when-piping-to-tee . Would it be possible to add `unbuffer` from the `expect` package so that we see normal live output again?

@marcoshuck

nkoenig:
See #947.

PR #947, opened 2021-06-02
Added unbuffer to the bridge and sim launch commands

nkoenig:
Resolves issue #945.

### DIFF
--- a/docker/cloudsim_bridge/run_bridge.bash
+++ b/docker/cloudsim_bridge/run_bridge.bash
@@ -21,4 +21,4 @@ export ROSCONSOLE_STDOUT_LINE_BUFFERED=1
 # atop process monitoring
 atop -R -w /home/developer/.ros/atop_log &
 
-ign launch cloudsim_bridge.ign -v 4 "$@" 2>&1 | tee /home/developer/.ros/bridge.log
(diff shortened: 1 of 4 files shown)


peci1:
The unbuffer call has to go in front of the `ign launch` command, not in front of the `tee`.

docker/cloudsim_bridge/run_bridge.bash
```diff
@@ -21,4 +21,4 @@ export ROSCONSOLE_STDOUT_LINE_BUFFERED=1
 # atop process monitoring
 atop -R -w /home/developer/.ros/atop_log &
 
-ign launch cloudsim_bridge.ign -v 4 "$@" 2>&1 | tee /home/developer/.ros/bridge.log
+unbuffer -p ign launch cloudsim_bridge.ign -v 4 "$@" 2>&1 | tee /home/developer/.ros/bridge.log
```
iche033, line 24: I think you don't need the `-p` if the `unbuffer` command is at the front, i.e. not after the pipe

nkoenig:
> The unbuffer call has to go in front of the `ign launch` command, not in front of the `tee`.

I think it's not fixed.

docker/cloudsim_bridge/run_bridge.bash
nkoenig: Thanks. It has been removed.

iche033 (approved):

PR #968, opened 2021-07-05
Mapping server transport dynamic cloud downsampling

acschang:
This PR includes a new node to decimate team cloud data by adjusting leaf node size in the PCL library filters.  Data is incrementally decimated to prevent submitted clouds having over 600,000 points to exceed the protobuf limits of Ignition transport resulting in the pointcloud being truncated.  The original size of the pointcloud prior to decimation is passed to the mapping server.

The node is added to the common launch file for mapping server relays and the relevant PCL dependencies are added to the requisite Dockerfiles.

### DIFF
--- a/docker/cloudsim_bridge/Dockerfile
+++ b/docker/cloudsim_bridge/Dockerfile
@@ -79,11 +79,14 @@ VOLUME /root/.aws
 RUN sudo /bin/sh -c 'echo "deb [trusted=yes] http://packages.ros.org/ros/ubuntu $(lsb_release -sc) main" > /etc/apt/sources.list.d/ros-latest.list' \
  && sudo apt-get update \
  && sudo apt-get install -y \
+    libpcl-dev \
+    libpcl-conversions-dev \
     python-catkin-tools \
     python-rosdep \
     python-rosinstall \
     ros-melodic-desktop \
     ros-melodic-joystick-drivers \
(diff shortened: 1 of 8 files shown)


iche033 (approved):
docker image built fine. I tested republishing point clouds from one of the robots to the `/cloud` topic and verified that the point clouds are downsampled in the `cloud_throttled_downsampled` topic

rsawtell:
To clarify, this can potentially change the size of pointclouds (eg lidar scans etc.) received by solutions in a fully dynamic way such that anything that expects a specific formatting of the pointcloud would break (ie: assuming an rgbd sensor always has 640x480 resolution) and that this happens when all the scans over all robots on a team exceeds a threshold?

acschang:
> To clarify, this can potentially change the size of pointclouds (eg lidar scans etc.) received by solutions in a fully dynamic way such that anything that expects a specific formatting of the pointcloud would break (ie: assuming an rgbd sensor always has 640x480 resolution) and that this happens when all the scans over all robots on a team exceeds a threshold?

This is accurate however the deployment of this node will only affect cloud data sent from teams to the mapping server so this will not affect the functionality of your solutions in any manner.  This node will affect the density of pointclouds sent to the mapping server as to stay below the 64MB transport limit.

More information on the mapping server is available on the [wiki](https://github.com/osrf/subt/wiki/api#mapping-server).

PR #981, opened 2021-07-20
Trigger the debris wall when it's empty of robots

nkoenig:
This will cause the debris wall to trigger only when empty of robots.


iche033:
also update `performer_detector_debris_wall_2` in finals_practice_02.sdf?

PR #992, opened 2021-08-12
Final prelim update

nkoenig:
This contains the final prelim worlds, and a couple small patches that:

1. Fixes the region detection event to output the correct type.
2. Adds logging information when `finish` is manually triggered.


iche033 (approved):
looks good to me

mjcarroll (approved):

PR #1009, opened 2021-09-22
Final event worlds 1-8

nkoenig:
This includes the first three final event worlds, and updates to the path tracer so that it can be run headless.


peci1:
This doesn't build on Dome since `msg.mutable_run_to_sim_time()` doesn't exist in ign-msgs6.

nkoenig:
This PR now has all of the final prize round worlds.

PR #1014, opened 2021-09-30
Compute gas concentration in ppm

nkoenig:
I'm copying over work done by @iche033 that provides a `ppm` component to the gas sensor.

A few changes are made to the gas detector. ~~The main breaking change is that the gas detector now publishes gas concentration (ppm) values instead of boolean true/false values.~~. **Update**: It now publishes both bool and double (ppm) values.

Other changes include simulating: 
* gas detector response time
* gas concentration level (based on distance to emitter source which is assumed to be center of room)
* Gaussian noise

The new simulated gas sensor device is based on Sensirion SCD30:
https://www.sensirion.com/en/environmental-sensors/carbon-dioxide-sensors/carbon-dioxide-sensors-scd30/

key parameters extracted from its spec:
* response time: 20s (t63%) 
* accuracy: +- (30ppm + 3%)
* range: 400 - 10,000 ppm

The new gas detector includes a few new configurable params:
* ~~`<legacy_detection_mode>`: True to publish boolean values, false to publish gas concentration (ppm) values. Defaults to false.~~
* `<response_time>`: Gas detector response time. Defaults to 20
  * This particular sensor device has a response time of 20s (t63%) response time (time it takes to detect 63% of an instance change in gas quantity). Actual response time may be longer.
* `<min_concentration>` : min gas  concentration level. Defaults to 412
* `<max_concentration>` : defaults to unbounded
* `<gas_concentration>`: target gas artifact concentration. Defaults to 1500 ppm
* `<precision>`: gas detector precision. Defaults to 30ppm

**To test:**

launch finals_qual world:

```
ign launch -v 4 competition.ign worldName:=finals_qual circuit:=finals robotName1:=X1 robotConfig1:=X1_SENSOR_CONFIG_1
```

In one terminal, echo the `gas_detected/ppm` topic:

```
rostopic echo /X1/gas_detected/ppm
```

You should see that the values are roughly around 410 ppm, which is the average CO2 concentration level in Earth atmosphere.  

Teleport the robot to one of the urban tiles that contains a gas artifact:

```
ign service -s /world/finals_qual/set_pose --reqtype ignition.msgs.Pose --reptype ignition.msgs.Boolean --timeout 300 --req 'name: "X1", position {x:80 y:105.5 z:10.0}'
```

You should then see that the gas concentration level slowly starts to rise until it's roughly hovering above 1500.

### DIFF
--- a/subt_ign/src/GasEmitterDetectorPlugin.cc
+++ b/subt_ign/src/GasEmitterDetectorPlugin.cc
@@ -18,6 +18,7 @@
 #include "subt_ign/GasEmitterDetectorPlugin.hh"
 
 #include <ignition/msgs/boolean.pb.h>
+#include <ignition/msgs/double.pb.h>
 
 #include <ignition/common/StringUtils.hh>
 #include <ignition/common/Console.hh>
@@ -63,6 +64,9 @@ class subt::GasDetectorPrivate
   /// \brief publisher to publish detections
   public: transport::Node::Publisher pub;
 
+  /// \brief publisher to publish ppm
+  public: transport::Node::Publisher ppmPub;
+
   /// \brief update rate
   public: double updateRate;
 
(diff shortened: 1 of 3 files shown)


iche033 (approved):
looks good to me. I only updated the PR description to describe the new behavior about supporting both bool and double gas detection outputs.

PR #1016, opened 2021-10-05
Nkoenig/localwindplugin

nkoenig:
The `fortress` branch is based on Ignition Fortress. You will need to the [melodic-fortress ros_ign branch](https://github.com/ignitionrobotics/ros_ign/tree/melodic-fortress), and set `IGNITION_VERSION=fortress` when running catkin_make. For example `IGNITION_VERSION=fortress catkin_make_install`.

This PR has copied the `finals_qual` world to a `finals_qual_demo` world, which will be used to demonstrate features added to SubT after the final event. 

The first such demonstration is a local wind model. The `LocalWindPlugin` allows a world to set the wind magnitude on a per-tile basis. I've added local wind to a few cave tiles located near the staging area. Specifically, tiles 28, 29, 37, and 38.

It's easiest to see the effect by increasing the magnitude to a value of 8 or higher, then move a robot into a cave tile. The wind will push the robot.


peci1:
Wouldn't this implementation allow "exploitation" of the simulation by simplifying detection of tile edges? Until now, it's been difficult once the robots left direct comms with teambase (with direct comms, tile edges could be estimated by large drops in rssi). 

What about "smoothing" the wind values by some gradual muxing function with neighbor tile wind speeds?

nkoenig:
> Wouldn't this implementation allow "exploitation" of the simulation by simplifying detection of tile edges? Until now, it's been difficult once the robots left direct comms with teambase (with direct comms, tile edges could be estimated by large drops in rssi).
> 
> What about "smoothing" the wind values by some gradual muxing function with neighbor tile wind speeds?

Interesting. I hadn't thought about that. I'll take another look. Thanks for the suggestion.

PR #1034, opened 2021-10-29
Record the teambase in the run.yml file

nkoenig:

The `run.yml` files did not record the presence of a TEAMBASE model because the teambase model lacks a sensor which bypassed the ECM `Each` call used to record robot names and types. 

The PR renames a link in the teambase to `teambase_link` and adds a check in the GameLogic plugin for the correct link name and also lack of a filepath. A run.yml file will now record the TEAMBASE model.


iche033 (approved):
works for me.

Issue #888, opened 2021-04-13
Strange LIDAR returns for various robots


rsawtell:
Looking through some of the robots, I've noticed that the LIDAR returns are sometimes wonky, typically but not always if part of the scan intersects robot geometry. I haven't exhaustively checked them all but below are a few for which it is particularly noticeable. For the UGV's the self intersecting scan is somehow displaced from the actual robot geometry, and for the UAV's there is a weird box shaped thing going on.

CSIRO_DATA61_DTR
![DTR](https://user-images.githubusercontent.com/700223/114549216-bee46580-9c2e-11eb-8ea3-3e0700c6b89c.png)

EXPLORER_R2
![explorer_r2](https://user-images.githubusercontent.com/700223/114549291-d1f73580-9c2e-11eb-82f5-7b1d90a29d05.png)

MARBLE_QAV500:
![qav500](https://user-images.githubusercontent.com/700223/114549403-f81cd580-9c2e-11eb-860c-e27c759fd6e1.png)

CERBERUS_Gagarin:
![gagarin](https://user-images.githubusercontent.com/700223/114549430-ff43e380-9c2e-11eb-8443-90620ccc69fa.png)

nkoenig:
The boxes seen in the gagarin and qav500 are effect of the min range associated with the laser sensors, which are currently set at 0.8 meters.

The returns for EXPLORER_R2 match the robot model in simulation. The robot description seen in rviz is not an accurate reflection of the physical robot.

The strange returns for CSIRO_DATA61_DTR are also related to a min range of 0.3 meters.

rsawtell:
Is the minimum distance using L1 norm then? I would expect default distance vectors to form a spherical pattern and not a box.

nkoenig:
I left out an important piece of information. The lidar is simulated using a camera with cube maps. We set the camera near clip distance to the laser minimum range. The effect is the box pattern you see. 

A fix would require changing the internal camera implementation to use a smaller near-clip and then clip the ranges after the GPU rendering process.

malcolmst:
I was going to ask the same question about the Marble HD2 lidar returns, but I think that answers it. Thanks for the info!

![marble hd2 lidar](https://user-images.githubusercontent.com/23696107/114956993-ac804c80-9e25-11eb-89a4-9c3a79eb850f.png)

peci1:
Can't it at least partly be caused by https://github.com/ignitionrobotics/ign-sensors/issues/128 ?

[...]

PR #1047, opened 2022-01-11
Model submission for Husky sensor config 3, 4 by CTU-CRAS-Norlab.

peci1:
Sensor configs 3, 4 are the CTU-CRAS build used in Finals. Sensor configs 1, 2 (already submitted and accepted) are the Norlab build that was a backup for the Finals and was not used in the end.

The motion characteristics should remain the same as sensor config 1 (just total mass is lower by 3 kg). You can check that comparing the relevant parts of model.sdf. I also fixed a weird inertia of the base link.

The sensors are the same used on MARV, Absolem 6 and other robots.

However, I did some tweaks to the Ouster model:

1. I decreased horizontal resolution to 1024 points/revolution as the 2048 mode is impossible to cool on the real robot and the lidar goes to shutdown very quickly.
2. I increased min range of the lidar to 25 cm. This corresponds to reality (they advertise 0 cm min range, but that is a lie/promise of a not yet released feature). This also prevents the lidar from outputting the weird artifacts observed e.g. in https://github.com/osrf/subt/issues/888#issuecomment-853357896 . I'll investigate if this isn't the simple way to fix Ousters on all of our robots without needing to dig into OGRE or changing lidar geometry assets.

I manually checked the look of collisions and inertia and they look okay. The inertia of sensor config 1's base link was weirdly rotated, so I fixed it in sensor configs 3,4. But maybe it is something to check for all the other Husky models. Cameras look okay. Lidar looks okay. Breadcrumbs are not present in config 3, and are usable with config 4. They fall from a place close to the dropper (it is intentionally in front of the robot). Battery drains. ROS-Ign bridge was checked for sensors, driving and breadcrumbs. Driving behaves as expected.


peci1:
@azeey Friendly ping.

